PR #2, opened 2023-11-13
reorg the script dir

skeeey:
reorganize the scripts dir to avoid the main redeclared in this block error when we use an IDE to edit the code


skeeey:
assign @clyang82

clyang82:
I was thinking that we should not have this `scripts` folder in maestro repo since it is cloned from trex. It should exist in trex only. @markturansky WDYT?

markturansky:
thank you, too, for keeping changes in sync with our "upstream".

https://github.com/openshift-online/rh-trex/pull/20/files

skeeey:
@markturansky thanks for making this clear, so I will remove these scripts in the maestro repo and only leave the generator in our upstream, does it make sense?

clyang82:
please also remove relevant files in templates folder started with `generate-`.

clyang82:
Yes. we need to exclude it when run the `clone` in TRex. And `clone` command should not exist in maestro repo. I will send a PR to fix it.

clyang82 (approved):

markturansky:
this PR should not have been merged.  as I wrote above: 

> we need the generate- templates, as they work with scripts/generator.go which is our local Kind generator.

This scripts/templates are useful to create Kinds in the future.

skeeey:
we removed the `scripts/generator.go` in this maestro repo, so my understanding is the `scripts/templates` will be also no longer needed

Do you mean we still need keep both of `scripts/generator.go` and `scripts/templates`  for future?

PR #24, opened 2023-12-13
Make e2e test

clyang82:
You can run `make e2e-test` to run the e2e tests. it includes deploy the maestro and maestro-agent in a KinD cluster and then do CRUD test for a resource.

PR #31, opened 2024-01-05
fix image pulling for agent deploy.

morvencao:
This PR tries to fix image pulling issue when deploying agent.


clyang82:
It was fixed in https://github.com/openshift-online/maestro/pull/24.

PR #34, opened 2024-01-11
fix maestro agent klog flags.

morvencao:
This pull request addresses the issue where agent klog flags are not functioning correctly due to conflicting definitions with glog(server) and klog in Maestro. The problem arises when both are used simultaneously, leading to the redefinition of flags and rendering agent flags ineffective. To resolve this, the PR introduces klog flags to the agent subcommand and implements a check to prevent redefinition before adding them to the flag set.

Now we can set `-v=5` to maestro agent:

```bash
...
I0111 03:36:58.747559       1 baseclient.go:120] Sent event: context.TODO
Context Attributes,
  specversion: 1.0
  type: io.open-cluster-management.works.v1alpha1.manifests.spec.resync_request
  source: b8da4ca7-052a-4322-8e12-5e38e82c5fce-work-agent
  id: bc1afd0e-d089-4d52-ba15-a03d78bff219
  time: 2024-01-11T03:36:58.74748891Z
  datacontenttype: application/json
Extensions,
  clustername: b8da4ca7-052a-4322-8e12-5e38e82c5fce
Data,
  {
    "resourceVersions": []
  }
  ...
```

/cc @clyang82 @skeeey

### DIFF
--- a/cmd/maestro/agent/cmd.go
+++ b/cmd/maestro/agent/cmd.go
@@ -14,6 +14,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	utilflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/version"
+	"k8s.io/klog/v2"
 	ocmfeature "open-cluster-management.io/api/feature"
 	"open-cluster-management.io/ocm/pkg/common/options"
 	"open-cluster-management.io/ocm/pkg/features"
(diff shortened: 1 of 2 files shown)


skeeey:
we will have two different log libraries glog and klog? could we consider to use an unified log library, e.g. the klog?

clyang82 (approved):
LGTM

maestro agent is running in a Kubernetes environment, using klog makes sense for me. for maestro, it can be running w/o Kubernetes environment. I think we do not need to consolidate them. Keep as it is.

PR #42, opened 2024-01-29
fix postgres permission

clyang82:
fix the issue `chmod: changing permissions of '/var/lib/postgresql/data': Operation not permitted` so that the postgres pod cannot be started in OCP environment.


morvencao (commented):
/lgtm
thanks!

qiujian16 (approved):

PR #64, opened 2024-03-21
Make integration test stable

clyang82:
1. do not need to re-create DB for each test. Just cleanup the table contents instead.
2. fix `concurrent map writes` by adding sync lock
As per golang doc:
>  Map access is unsafe only when updates are occurring. As long as all goroutines are only reading—looking up elements in the map, including iterating through it using a for range loop—and not changing the map by assigning to elements or doing deletions, it is safe for them to access the map concurrently without synchronization.
3. adopt fix https://github.com/open-cluster-management-io/sdk-go/pull/26

### DIFF
--- a/pkg/db/advisory_locks.go
+++ b/pkg/db/advisory_locks.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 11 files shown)


pkg/db/advisory_locks.go
```diff
@@ -93,7 +101,9 @@ func (f *AdvisoryLockFactory) NewNonBlockingLock(ctx context.Context, id string,
 	}
 
 	log.V(4).Info(fmt.Sprintf("Locked non blocking advisory lock id=%s type=%s - owner=%s", id, lockType, *lock.uuid))
+	f.mutex.Lock()
```
skeeey, line 104: we may make the `advisoryLockMap` thread safe, and we also need commit this to the upstream

gurnben (approved):
/lgtm
/approve

PR #74, opened 2024-04-01
adds missing serviceAccountName to deployment template

markturansky:
adds missing serviceAccountName to deployment template


gurnben (approved):
lgtm!

Issue #67, opened 2024-03-25
Introduce "external_id"


clyang82:
if you post the same resource, i believe you'd get dupes.   Maestro isn't inspecting the generic payload.  not sure this is even practical.
it might be best, then, to include an "external_id" that CS can provide (its own PK for that record, for example) that is unique.
in this way:
TX1: CS persists to db, gets a PK
TX2: CS posts resource w/ the PK.   it would fail if the same resource is posted more than once.
TX3: Maestro can have many copies of this resource in the queue.  resources are applied idempotently in the cluster.

FYI @markturansky

clyang82:
working on this issue.

markturansky:
@clyang82 is this a moot point now that `name` is provided by the caller and enforced for uniqueness?

markturansky:
see https://github.com/openshift-online/maestro/pull/63

clyang82:
Yes. I think it can be. if the name is provided, then we use it and enforce for uniqueness. if the name is not provided, then we use ID as the name.

PR #78, opened 2024-04-16
Add resource name as an external id

clyang82:
fixed: https://github.com/openshift-online/maestro/issues/67


openapi/openapi.yaml
```diff
@@ -42,6 +42,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
       parameters:
+      - $ref: '#/components/parameters/name'
```
skeeey, line 45: we may not need to introduce this parameter, it seems we can get resource with search parameter, e.g.
`/api/maestro/v1/resources?search=name = 'cluster3'`

clyang82: yes. search can be used for any field name. But it seems `/api/maestro/v1/resources?name=cluster3` is widely used.

pkg/services/types.go
```diff
@@ -15,6 +15,7 @@ type ListArguments struct {
 	Search   string
 	OrderBy  []string
 	Fields   []string
+	Name     string
```
morvencao, line 18: why does list argument contain `name`? can it just replace with search?

openapi/openapi.yaml
clyang82: @markturansky How do you use in your side?

clyang82:
remove query by name and  in favor of `ocm get /api/maestro/v1/resources --parameter search="name = 'nginx'"`

pkg/services/resource.go
```diff
@@ -56,6 +56,11 @@ func (s *sqlResourceService) Get(ctx context.Context, id string) (*api.Resource,
 }
 
 func (s *sqlResourceService) Create(ctx context.Context, resource *api.Resource) (*api.Resource, *errors.ServiceError) {
+	if resource.Name != "" {
+		if err := ValidateResourceName(resource); err != nil {
+			return nil, errors.Validation("the name in the resource is invalid, %v", err)
+		}
+	}
```
qiujian16, line 63: should we also check the name immutable when update?

clyang82: Name string `gorm:"uniqueIndex"` there is error from database. we do not need to check here again for updating.

qiujian16 (approved):

pkg/services/resource.go
clyang82: the name is not exposed for patch. so the name will be ignored if it is changed when patch.

PR #91, opened 2024-05-17
docs: add a Why section to the readme

mikeshng:
I am trying to show maestro as a possible solution for another project but the README is missing some motivations on why this project is needed. This PR is to address that missing overview.


README.md
```diff
@@ -1,12 +1,40 @@
...
 
 ## Architecture
+
 Taking MQTT as an example:
+
 ![Maestro Architecture](./arch.png)
 
+## Why
+
+Maestro was created to address several critical challenges associated with managing Kubernetes clusters at scale.
```
markturansky, line 15: This paragraph is good in that we should see performance and scalability improvements by using Maestro, as you note.

Another motivation (I think, perhaps, primary motivation) was to improve the security architecture.

By using Maestro, we do not need kubeconfigs.  Our Cluster Service (CS) deployment would no longer need kubeconfigs or direct access to customer clusters. CS just uses the Maestro API. 

Only the Maestro consumer in a management cluster needs kubeconfigs, thus improving security.

mikeshng: Thanks Mark for the review!

Could you please help me word it in a way that's more generic and not mention CS, SD, MC, HCP etc?
I am trying to show maestro as a possible solution for another project and it's not related to SD. Maybe closer to self managed?

mikeshng: Added a generic sentence about not needing kubeconfig. Maybe it's a bit vague but PTAL. Thanks.

clyang82:
@mikeshng Thanks for your PR. Maybe add a description for the purpose of this PR. something like show maestro as a possible solution for another project.

qiujian16:
lgtm

clyang82:
/lgtm

leave it to @markturansky for another look.

clyang82 (approved):

PR #94, opened 2024-05-21
expose kubeconfig flag to run the agent binary on a host

skeeey:


cmd/maestro/agent/cmd.go
```diff
@@ -47,53 +58,98 @@ func NewAgentCommand() *cobra.Command {
...
-	cfg := spoke.NewWorkAgentConfig(commonOptions, agentOption)
-	cmdConfig := commonOptions.CommoOpts.
-		NewControllerCommandConfig("maestro-agent", version.Get(), cfg.RunWorkloadAgent)
-	cmdConfig.DisableLeaderElection = true
-
-	if err := cmdConfig.StartController(ctx); err != nil {
-		glog.Fatalf("error running command: %v", err)
+	workOptions.WorkloadSourceDriver = "mqtt"
+	// use manifest as the default codec
+	workOptions.CloudEventsClientCodecs = []string{"manifest"}
```
clyang82, line 83: Could you update to support `manifest` and `manifestbundle` by default?

cmd/maestro/agent/cmd.go
```diff
@@ -47,53 +58,98 @@ func NewAgentCommand() *cobra.Command {
...
+	CommonOptions  *options.AgentOptions
+	WorkOptions    *spoke.WorkloadAgentOptions
+	KubeConfigFile string
+	Namespace      string
+}
 
+func NewAgentOptions() *AgentOptions {
+	workOptions := spoke.NewWorkloadAgentOptions()
+	// use 1M as the default limit for state feedback
+	workOptions.MaxJSONRawLength = 1024 * 1024
```
clyang82, line 79: I remembered it was set in agent options. Do you need to reset it again?

skeeey: yes, but the default value is small for maestro case, we need increase it here

cmd/maestro/agent/cmd.go
skeeey: I'm not sure, we should enable both codecs at the same time, for maestro, it should only handle the manifest, right?

clyang82: Let us handle it in separate issue https://github.com/openshift-online/maestro/issues/96

cmd/maestro/agent/cmd.go
```diff
@@ -47,53 +58,98 @@ func NewAgentCommand() *cobra.Command {
...
-		agentOption.CloudEventsClientID, "The ID of the agent client, by default it is <consumer-id>-work-agent")
-	fs.StringSliceVar(&agentOption.CloudEventsClientCodecs, "agent-client-codecs",
-		[]string{"manifest"}, "The codecs of the agent client. The valid codecs are manifest and manifestbundle")
+	fs.StringVar(&o.WorkOptions.WorkloadSourceConfig, "message-broker-config-file",
+		o.WorkOptions.WorkloadSourceConfig, "The config file path of the message broker, it can be mqtt broker or kafka broker")
+	fs.StringVar(&o.WorkOptions.WorkloadSourceDriver, "message-broker-type",
+		o.WorkOptions.WorkloadSourceDriver, "Message broker type")
+	fs.StringVar(&o.WorkOptions.CloudEventsClientID, "agent-client-id",
+		o.WorkOptions.CloudEventsClientID, "The ID of the agent client, by default it is <consumer-id>-work-agent")
+	fs.StringSliceVar(&o.WorkOptions.CloudEventsClientCodecs, "agent-client-codecs",
```
machi1990, line 145: can we propagate that up to the agent template https://github.com/openshift-online/maestro/blob/e3bdb59e5444b1e87b0a28828e0ae59d034f010c/templates/agent-template.yml ?

skeeey:
we need to use library-go to implement this to support lease election

Issue #69, opened 2024-03-27
Update and delete policy in resource API


qiujian16:
To support different update/delete strategy

PR #99, opened 2024-05-23
Support UpdateStrategy and DeleteOption

clyang82:
fixed: https://github.com/openshift-online/maestro/issues/69

Introduce to configure the `delete_option` and `update_strategy` when creating or patching resource.

delete_option defines the option to delete the resource. It is optional when creating a resource. The propagationPolicy of `delete_option` can be:
- `Foreground` represents that the resource should be fourground deleted. This is a default value.
- `Orphan` represents that the resource is orphaned when deleting the resource.

update_strategy defines the strategy to update the resource. It is optional when creating a resource. The type of `update_strategy` can be:
- `ServerSideApply` means to update resource using server side apply with work-controller as the field manager. This is a default value.
- `Update` means to update resource by an update call.
- `CreateOnly` means do not update resource based on current manifest.
- `ReadOnly` means only check the existence of the resource based on the resource's metadata.


pkg/handlers/resource.go
```diff
@@ -70,7 +70,7 @@ func (h resourceHandler) Patch(w http.ResponseWriter, r *http.Request) {
 		func() (interface{}, *errors.ServiceError) {
 			ctx := r.Context()
 			id := mux.Vars(r)["id"]
-			manifest, err := presenters.ConvertResourceManifest(patch.Manifest)
+			manifest, err := presenters.ConvertResourceManifest(patch.Manifest, patch.DeleteOption, patch.UpdateStrategy)
```
qiujian16, line 73: I think we should support update these options.

pkg/api/openapi/model_resource.go
```diff
@@ -406,6 +472,12 @@ func (o Resource) ToMap() (map[string]interface{}, error) {
 	if !IsNil(o.Manifest) {
 		toSerialize["manifest"] = o.Manifest
 	}
+	if !IsNil(o.DeleteOption) {
+		toSerialize["delete_option"] = o.DeleteOption
```
qiujian16, line 476: delete_option or deleteOption?

pkg/api/resource_types.go
```diff
@@ -121,19 +121,48 @@ func CloudEventToJSONMap(evt *cloudevents.Event) (datatypes.JSONMap, error) {
...
+
 	// create a cloud event with the manifest as the data
 	evt := cetypes.NewEventBuilder("maestro", cetypes.CloudEventsType{}).NewEvent()
 	eventPayload := &workpayload.Manifest{
-		Manifest: unstructured.Unstructured{Object: manifest},
-		DeleteOption: &workv1.DeleteOption{
-			PropagationPolicy: workv1.DeletePropagationPolicyTypeForeground,
-		},
+		Manifest:     unstructured.Unstructured{Object: manifest},
+		DeleteOption: delOption,
```
qiujian16, line 165: @skeeey would you confirm deleteOption here works in agent side?

pkg/api/openapi/model_resource.go
clyang82: in restful api, it seems we should use `delete_option`. for example: `consumer_name` `create_at` etc https://github.com/openshift-online/maestro?tab=readme-ov-file#get-your-resource

pkg/handlers/resource.go
clyang82: Do you mean `PATCH` these options? if so, it was covered. I am working on adding test cases for that.

pkg/api/resource_types.go
skeeey: what kinds of cases are you concerned about?
we already have some test in our integration test https://github.com/open-cluster-management-io/ocm/blob/main/test/integration/cloudevents/deleteoption_test.go

qiujian16 (approved):

PR #100, opened 2024-05-27
deploy maestro with docker option

yanmxa:
I try to deploy the meastro to the openshift cluster. It throws the following error message:
```bash
make deploy
for cmd in $(ls cmd); do \
        go build \
                -ldflags="" \
                -o "${cmd}" \
                "./cmd/${cmd}" \
                || exit 1; \
done
podman build -t "default-route-openshift-image-registry.apps.obs-hub-of-hubs-aws-414-sno-lrqtk.scale.red-chesterfield.com/maestro-myan/maestro:1716775881" .
/bin/sh: line 1: podman: command not found
make: *** [Makefile:314: image] Error 127
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,8 @@ environment:=${USER}
 namespace ?= maestro-${USER}
 agent_namespace ?= maestro-agent-${USER}
 
-# a tool for managing containers and images, etc. You can set it as docker
-container_tool ?= podman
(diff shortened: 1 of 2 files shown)


yanmxa:
Currently, we can explicitly use the container tool to set the setting. We could enable automatic detection in the future.

PR #108, opened 2024-05-31
add consumer test for e2e

yanmxa:

Add the e2e test case for CURD on consumers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -410,9 +410,10 @@ e2e-test/teardown:
 .PHONY: e2e-test/teardown
 
 e2e-test: e2e-test/teardown e2e-test/setup
-	ginkgo -v --output-dir="${PWD}/test/e2e/report" --json-report=report.json --junit-report=report.xml \
-	${PWD}/test/e2e/pkg -- -consumer_name=$(shell cat ${PWD}/test/e2e/.consumer_name) \
+	ginkgo --output-dir="${PWD}/test/e2e/report" --json-report=report.json --junit-report=report.xml \
+	${PWD}/test/e2e/pkg -- \
 	-api-server=https://$(shell cat ${PWD}/test/e2e/.external_host_ip):30080 \
 	-grpc-server=$(shell cat ${PWD}/test/e2e/.external_host_ip):30090 \
-	-kubeconfig=${PWD}/test/e2e/.kubeconfig
+	-consumer-name=$(shell cat ${PWD}/test/e2e/.consumer_name) \
+	-consumer-kubeconfig=${PWD}/test/e2e/.kubeconfig
 .PHONY: e2e-test
--- a/test/e2e/pkg/consumer_test.go
+++ b/test/e2e/pkg/consumer_test.go
@@ -1,0 +1,72 @@
+package e2e_test
+
+import (
+	"net/http"
+	"reflect"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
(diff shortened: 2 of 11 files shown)


test/e2e/pkg/suite_test.go
```diff
@@ -19,40 +20,32 @@ import (
 )
 
 var (
+	consumerOpts     *ConsumerOptions
 	apiServerAddress string
-	kubeconfig       string
-	consumer_name    string
-	kubeClient       *kubernetes.Clientset
 	apiClient        *openapi.APIClient
 	helper           *test.Helper
```
morvencao, line 26: @clyang82 
I'm curious why we need `helper` in e2e testing? It's mainly used by integration test to set up maestro servers, but for e2e testing, the only use of `helper` is to create openapi resource?

Makefile
```diff
@@ -406,8 +406,14 @@ e2e-test/teardown:
 	./test/e2e/setup/e2e_teardown.sh
 .PHONY: e2e-test/teardown
 
-e2e-test: e2e-test/teardown e2e-test/setup
+e2e-test:
+	@if [ -z "$$consumer_name" ] || [ -z "$$api_server" ] || [ -z "$$consumer_kubeconfig" ]; then \
```
clyang82, line 410: why need this judgement?

test/e2e/pkg/suite_test.go
clyang82: Yes. only use it to create openapi resource.

test/e2e/pkg/consumer_test.go
```diff
@@ -0,0 +1,72 @@
...
+			Expect(consumerList).NotTo(BeNil())
+			Expect(len(consumerList.Items) > 0).To(BeTrue())
+
+			got := false
+			for _, consumer := range consumerList.Items {
+				if *consumer.Name == consumerOpts.Name {
+					got = true
+				}
+			}
+			Expect(got).To(BeTrue())
```
clyang82, line 34: how to guarantee the consumer is there? the test order is random. if run this test at the beginning, the consumer should be example. right?

Makefile
yanmxa: By default, the e2e test will run on the KinD cluster by running `e2e-test/setup`. Here we can provide a way to run the  `e2e-test` on other platforms like `AKS` or `Openshift`. All we need to do is given the environment variables `consumer_name`, `api_server`, and `consumer_kubeconfig`.

I found it convenient during testing, but could this be considered over-engineering? Do you have any suggestions?

test/e2e/pkg/consumer_test.go
yanmxa: For this case, we already have generated a consumer before running the e2e test. So the consumer must be there~

Makefile
clyang82: maybe export a single environment to indicate whether need to setup the environment or use the existing environment. What do you think?

test/e2e/pkg/consumer_test.go
clyang82: How about the case you want to run it in an existing environment? Do you require create the consumer as a pre-requirement? maybe just adjust your order to create/get/patch

clyang82 (changes requested):

Makefile
yanmxa: Maybe we can add this feature in another PR, I removed it from the current.

clyang82 (approved):
LGTM

PR #110, opened 2024-06-04
rename manifest to payload in database

clyang82:


qiujian16 (approved):

pkg/db/migrations/202311151850_add_resources.go
```diff
@@ -18,8 +18,8 @@ func addResources() *gormigrate.Migration {
 		// "Single" resource type for RESTful API calls,
 		// "Bundle" resource type mainly for gRPC calls.
 		Type string `gorm:"index"`
-		// Manifest holds the resource manifest in CloudEvent format (JSON representation).
-		Manifest datatypes.JSON `gorm:"type:json"`
+		// Payload is CloudEvent payload with CloudEvent format (JSON representation).
+		Payload datatypes.JSON `gorm:"type:json"`
```
machi1990, line 22: FYI 

I am seeing this 
```
failed to create resource: maestro-9: Unable to create Resource: pq: column "payload" of relation "resources" does not exist
```

when I deploy the latest `main` against an existing deployment.
Ideally, this should've been a new migration file so that the column rename is applied. Renaming a column here, won't work for existing deployments as the migration id has already been applied.

Issue #101, opened 2024-05-28
Creating a resource with the same name after its deletion fails


machi1990:
Reproducer:

1. Create the resource
```
ocm post /api/maestro/v1/resources << EOF
{
        "consumer_name": "cluster1", "name": "mambo-2",
        "version": 1,
        "manifest": {
          "apiVersion": "work.open-cluster-management.io/v1",
          "kind": "ManifestWork",
          "metadata": {
                "name": "mw",
                "namespace": "cluster1"
          },
          "spec": {
                "workload": { "manifests": [
                        {
                          "apiVersion": "v1",
                          "data": {
                                "mambo": "poa"
                          },
                          "kind": "Secret",
                          "metadata": {
                                "name": "test",
                                "namespace": "test"
                          }
                        }
                  ]}
          }
        }
  }
EOF
```

2. Delete the resource using the id given after the create call; replace the `<resource_id>` below with the id returned from 1
```
ocm delete /api/maestro/v1/resources/<resource_id>
```
3. Once the resource is successful deleted, attempt to recreate it again using step (1) above. And observe the failure.

A quick debug reveals this to be a DB level constraint of uniqueness name and soft deletion.
Since the resource in step(2) is soft deleted, it's still present in the database. Recreating a resource with the same name afterwards fail since the resource still exists in the DB. 

The constraint is defined as 
```
 "idx_resources_name" UNIQUE, btree (name)
```

clyang82:
Do you have agent available? the agent should report back the deletion status and then we will delete from DB. We may introduce force deletion after 5 mins.

machi1990:
> the agent should report back the deletion status and then we will delete from DB

@clyang82 Hey, Yes. The agent is available. I confirm that deleted resource was marked for deletion successfully and then soft deleted. The issue is that once it was soft deleted, it remained in the database making it impossible to create another resource with the same name.

clyang82:
@machi1990 this issue was fixed. please verify it. Thanks.

machi1990:
Thank you. I've verified the code change. It looks good to me. I'll close the issue.

PR #111, opened 2024-06-05
enhance resource deletion flow.

morvencao:
This PR enhance the resource deletion flow by:
1. Soft delete the resource record when the request reach to Maestro Server
2. Publish delete cloud event to Maestro agent by setting deletionTimeStamp
3. Agent delete the resource from cluster and publish status update with `ManifestsDeleted` condition
4. Maestro check the `ManifestsDeleted` condition in the status update and delete the resource record permanently.


pkg/api/presenters/resource.go
```diff
@@ -61,7 +61,14 @@ func PresentResource(resource *api.Resource) (*openapi.Resource, error) {
 		DeleteOption:   deleteOption,
 		UpdateStrategy: updateStrategy,
 		Status:         status,
-	}, nil
+	}
+
+	// set the deletedAt field if the resource has been marked as deleted
+	if !resource.DeletedAt.Time.IsZero() {
+		res.DeletedAt = openapi.PtrTime(resource.DeletedAt.Time)
```
clyang82, line 68: why need this check? can we just use `DeletedAt: = openapi.PtrTime(resource.DeletedAt)`

morvencao: this should never happen, just for double check.
because 'onDelete' will only called after the `MarkAsDeleting`.

clyang82:
fixed: https://github.com/openshift-online/maestro/issues/101

clyang82 (approved):
LGTM

PR #113, opened 2024-06-06
fix maestro server template for ARO.

morvencao:
1. enable 'broadcast' subscription type for maestro server on ARO, because event grid MQTT server doesn't support shared subscription, see: https://learn.microsoft.com/en-us/azure/event-grid/mqtt-support#mqttv5-current-limitations
2. remove duplicated DB SSL mode argument entry.


clyang82 (approved):
LGTM

clyang82:
need rebase your PR to ensure the ut and integration pass.

morvencao:
rebased the code.

Issue #109, opened 2024-05-31
Test the maestro server and agent on AKS


yanmxa:
1. Add the e2e test case that can run on both KinD cluster and AKS

2. Delete resources failed on this [case](https://github.com/openshift-online/maestro/blob/253a30058b78cd8e9a4467d2393e769dadf8a1dd/test/e2e/pkg/resources_test.go#L109-L137)
    ```
        IMAGE_REGISTRY=quay.io \
        IMAGE_REPOSITORY=redhat-user-workloads/maestro-rhtap-tenant/maestro/maestro \
        IMAGE_TAG=0ba050b00ef480cf2da6bd937a0ca3711c51644d \
    ```
    
3. Not found the deletion API on the consumer 

4. Heartbeat failed when the server rerun again

PR #125, opened 2024-06-12
Failed to update server instance heartbeat to the database

yanmxa:
When the maestro server is down for a while, it's up again. It might not sync its heartbeat to the database causing the following reasons:
```bash
E0612 09:28:05.379536   72292 logger.go:129]   failed to mark transaction for rollback: could not retrieve transaction from context 
E0612 09:28:05.380320   72292 logger.go:129]   Unable to upsert maestro instance: pq: duplicate key value violates unique constraint "server_instances_pkey" 
```

So the behavior will be:
1. The server is up, update the heartbeat
2. The server is down, then the liveness goroutine will mark the server as deleted if it reaches a specific duration
3. The server is up again, it will update the the heartbeat to the database and then mark the `deleted_at` as null

Reference: https://github.com/openshift-online/maestro/issues/109


clyang82 (approved):
Good catch!

morvencao (commented):
/lgtm

Issue #102, opened 2024-05-29
Allow search a manifest by label


machi1990:
I've a manifest created via the restful API that looks like this
```json
{
      "consumer_name":"cluster1",
      "created_at":"2024-05-29T16:08:13.719615Z",
      "href":"/api/maestro/v1/resources/dd54c3b7-a0af-401b-a77e-2c9faf5395f7",
      "id":"dd54c3b7-a0af-401b-a77e-2c9faf5395f7",
      "kind":"Resource",
      "manifest": {
        "apiVersion":"apps/v1",
        "kind":"Deployment",
        "metadata": {
          "name":"nginx",
          "namespace":"default"
        },
        "spec": {
          "replicas":1,
          "selector": {
            "matchLabels": {
              "app":"nginx"
            }
          },
          "template": {
            "metadata": {
              "labels": {
                "app":"nginx"
              }
            },
            "spec": {
              "containers": [
                {
                  "image":"nginxinc/nginx-unprivileged",
                  "name":"nginx"
                }
              ]
            }
          }
        }
      }
      ....
```

Listing for it without a search parameter works,

```json
ocm get /api/maestro/v1/resources                                                                    
{
  "items": [
    {
      "consumer_name":"cluster1",
      "created_at":"2024-05-29T16:08:13.719615Z",
      "href":"/api/maestro/v1/resources/dd54c3b7-a0af-401b-a77e-2c9faf5395f7",
      "id":"dd54c3b7-a0af-401b-a77e-2c9faf5395f7",
      "kind":"Resource",
      "manifest": {
        "apiVersion":"apps/v1",
        "kind":"Deployment",
        "metadata": {
          "name":"nginx",
          "namespace":"default"
        },
        "spec": {
          "replicas":1,
          "selector": {
            "matchLabels": {
              "app":"nginx"
            }
          },
          "template": {
            "metadata": {
              "labels": {
                "app":"nginx"
              }
            },
            "spec": {
              "containers": [
                {
                  "image":"nginxinc/nginx-unprivileged",
                  "name":"nginx"
                }
              ]
            }
          }
        }
      },
      "name":"dd54c3b7-a0af-401b-a77e-2c9faf5395f7",
      "updated_at":"2024-05-29T16:08:13.719615Z",
      "version":1
    },
...

```

However, when I attempt to search for it using the label, I receive an error from Maestro 

```shell
[...]

PR #127, opened 2024-06-13
Support jsonb search

clyang82:
fixed: https://github.com/openshift-online/maestro/issues/102

### DIFF
--- a/pkg/services/generic.go
+++ b/pkg/services/generic.go
@@ -2,17 +2,18 @@ package services
 
 import (
 	"context"
+	"encoding/json"
 	e "errors"
 	"fmt"
 	"reflect"
 	"strings"
 
-	"gorm.io/gorm"
-
 	"github.com/Masterminds/squirrel"
 	"github.com/yaacov/tree-search-language/pkg/tsl"
 	"github.com/yaacov/tree-search-language/pkg/walkers/ident"
 	sqlFilter "github.com/yaacov/tree-search-language/pkg/walkers/sql"
+	"gorm.io/gorm"
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	"github.com/openshift-online/maestro/pkg/api"
 	"github.com/openshift-online/maestro/pkg/dao"
@@ -151,6 +152,16 @@ func (s *sqlGenericService) buildOrderBy(listCtx *listContext, d *dao.GenericDao
 func (s *sqlGenericService) buildSearch(listCtx *listContext, d *dao.GenericDao) (bool, *errors.ServiceError) {
 	if listCtx.args.Search == "" {
 		s.addJoins(listCtx, d)
+		return true, nil
+	}
+
+	if isJSONBSearch(listCtx.args.Search) {
+		// focus on jsonb, ignore type = 'Single' or 'Bundle'
+		jsonbSearch, _, _ := strings.Cut(listCtx.args.Search, "and type=")
+		if err := validateJSONBSearch(jsonbSearch); err != nil {
+			return false, errors.BadRequest("failed to validate search query: %v", err)
+		}
+		(*d).Where(listCtx.args.Search, nil)
 		return true, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/services/generic.go
```diff
@@ -154,6 +155,16 @@ func (s *sqlGenericService) buildSearch(listCtx *listContext, d *dao.GenericDao)
 		return true, nil
 	}
 
+	if isJSONBSearch(listCtx.args.Search) {
+		// focus on jsonb, ignore type = 'Single' or 'Bundle'
+		jsonbSearch, _, _ := strings.Cut(listCtx.args.Search, "and type=")
```
qiujian16, line 160: what does "and type=" mean?

pkg/services/generic.go
```diff
@@ -333,3 +344,66 @@ func (s *sqlGenericService) treeWalkForSqlizer(listCtx *listContext, tslTree tsl
...
+		return true
+	}
+	return false
+}
+
+func validateJSONBSearch(search string) error {
+	// handle ->> search. for example: payload -> 'data' -> 'manifest' -> 'metadata' -> 'labels' ->> 'foo' = 'bar'
+	if strings.Contains(search, "->>") {
+		jsonbStr := stringSplitwithTrim(search, "->>")
+		if err := isValidFields(stringSplitwithTrim(jsonbStr[0], "->")); err != nil {
```
qiujian16, line 359: do we ensure jsonbStr is always not nil

pkg/services/generic.go
```diff
@@ -333,3 +344,66 @@ func (s *sqlGenericService) treeWalkForSqlizer(listCtx *listContext, tslTree tsl
...
+			return fmt.Errorf("the search field name is invalid: %v", err)
+		}
+		labelStr := stringSplitwithTrim(jsonbStr[1], "=")
+		if err := validation.IsQualifiedName(labelStr[0]); err != nil {
```
qiujian16, line 363: ditto, do we ensure labelStr is alway not nil?

pkg/services/generic.go
```diff
@@ -333,3 +344,66 @@ func (s *sqlGenericService) treeWalkForSqlizer(listCtx *listContext, tslTree tsl
...
+		}
+	}
+	// handle @> search. for example: payload -> 'data' -> 'manifests' @> '[{\"metadata\":{\"labels\":{\"foo\":\"bar\"}}}]'
+	if strings.Contains(search, "@>") {
+		jsonbStr := stringSplitwithTrim(search, "@>")
+		if err := isValidFields(stringSplitwithTrim(jsonbStr[0], "->")); err != nil {
+			return fmt.Errorf("the search field name is invalid: %v", err)
+		}
+		// validate the value is a valid json object
+		if !isValidJSONObject(jsonbStr[1]) {
```
qiujian16, line 379: Do we support any json string?

pkg/services/generic.go
clyang82: it was added at the beginning to decide the type is `Single` or `Bundle`. refer to https://github.com/openshift-online/maestro/blob/7a3daed319b16acc20ee3a34f6a8e97c533e253e/pkg/handlers/resource.go#L104. Here just focus on jsonb search

pkg/services/generic.go
clyang82: Yes. support any json only if it is valid json object.

pkg/services/generic.go
clyang82: if the search string has "->>", it can enter this method. that means jsonbStr[0] and jsonbStr[1] is always not nil. for example: "'labels'->>", the jsonbStr[0] is `labels` and jsonbStr[1] is empty.

pkg/services/generic.go
qiujian16: what is there is no "="?

pkg/services/generic.go
qiujian16: what if only "->>"? Also what if there are two "->>" in it? Do we allow only one?

pkg/services/generic.go
qiujian16: do we need to support any json string? For instance, even a spec field in the json string?

pkg/services/generic.go
clyang82:
```
		{
			"name":   "invalid search without field name",
			"search": "->>",
			"error":  "the search field name is invalid",
		},
		{
			"name":   "invalid search with two ->> symbols",
			"search": "payload -> 'data' -> 'manifest' -> 'metadata' -> 'labels'->> 'foo' = 'bar' AND 'labels'->> 'foo'",
			"error":  "the search value is invalid",
		},

```
Added 2 tests to cover the above case. Yes. we allow only one `->>` symbol.

pkg/services/generic.go
clyang82: refer to this test
```
		{
			"name":   "complex labels",
			"search": "payload -> 'data' -> 'manifest' -> 'metadata' -> 'labels' ->> 'example.com/version'",
		},
```

qiujian16 (approved):

PR #128, opened 2024-06-14
Support set the limit when list resources by source work client 

skeeey:


machi1990:
lgtm

This should get us going with our use case.

pkg/client/cloudevents/grpcsource/watcherstore.go
```diff
@@ -107,8 +107,15 @@ func (m *RESTFulAPIWatcherStore) Get(namespace, name string) (*workv1.ManifestWo
...
 
+	var size int32 = -1
+	if opts.Limit > 0 {
+		size = int32(opts.Limit)
+	}
+
 	apiRequest := m.apiClient.DefaultApi.ApiMaestroV1ResourceBundlesGet(context.Background()).
-		Search(fmt.Sprintf("source = '%s'", m.sourceID))
+		Search(fmt.Sprintf("source = '%s'", m.sourceID)).
+		Page(1).
```
machi1990, line 117: I can see other people wanting to configure this as well - but currently the opts doesn't have it but rather a continue field. But it's something to keep in mind on how it could be supported later on.

skeeey: add a TODO for this

machi1990 (approved):

PR #133, opened 2024-06-18
support full sync when the work watcher creating or the work client reconnected

skeeey:
refer to https://github.com/openshift-online/maestro/issues/131

PR #135, opened 2024-06-19
support listing works with its labels

skeeey:


skeeey:
should wait for #133 is merged

pkg/client/cloudevents/grpcsource/util.go
```diff
@@ -86,6 +90,39 @@ func ToManifestWork(rb *openapi.ResourceBundle) (*workv1.ManifestWork, error) {
...
+
+	requirements, selectable := labelSelector.Requirements()
+	if !selectable {
+		return "", false, nil
+	}
+
+	labels := []string{}
+	for _, requirement := range requirements {
+		if requirement.Operator() != selection.Equals {
+			// TODO support more operators
```
machi1990, line 111: With https://github.com/openshift-online/maestro/pull/136 being merged, can we check what other operators are supported and allow list them as valid operators? 

/cc @clyang82

skeeey: @machi1990 now only the `DoesNotExist` cannot be supported (if this is expected, we need enhance the `ocm-sdk` as a follow up), I added some test case in the e2e `List works with gRPC source ManifestWork client`, and I also add the support for label selector in the watch, PTAL, thanks

machi1990: Thanks @skeeey 

>  now only the DoesNotExist cannot be supported (if this is expected, we need enhance the ocm-sdk as a follow up)

That's noted. Just out of curiosity, wouldn't his be the opposite of what's done for Exists? At least that's my expectation but maybe I am missing something.

> I added some test case in the e2e List works with gRPC source ManifestWork client, and I also add the support for label selector in the watch, PTAL, thanks

I am giving it a look now. Thanks

machi1990: The change lgtm. 
Once this is merged, I can test it against my client I am building.

machi1990:
lgtm

Once merged, I can test it against the client I am building.

pkg/client/cloudevents/grpcsource/util.go
```diff
@@ -86,6 +90,88 @@ func ToManifestWork(rb *openapi.ResourceBundle) (*workv1.ManifestWork, error) {
...
+
+			equalsLabels = append(equalsLabels, fmt.Sprintf(`"%s":"%s"`, requirement.Key(), values.List()[0]))
+		case selection.NotEquals:
+			values := requirement.Values()
+			if len(values) != 1 {
+				return nil, "", false, fmt.Errorf("too many values in not equals operation")
+			}
+
+			notEqualsLabels = append(notEqualsLabels, fmt.Sprintf(`payload->'metadata'->'labels'->>'%s'<>'%s'`,
+				requirement.Key(), values.List()[0]))
```
clyang82, line 135: maybe extract `payload->'metadata'->'labels'` as a constant since it is using in many places.

pkg/client/cloudevents/grpcsource/util.go
```diff
@@ -86,6 +90,88 @@ func ToManifestWork(rb *openapi.ResourceBundle) (*workv1.ManifestWork, error) {
...
+		for key, vals := range inLabels {
+			labelSearch = append(labelSearch, fmt.Sprintf(`payload->'metadata'->'labels'->>'%s'in(%s)`,
+				key, strings.Join(vals, ",")))
+		}
+	}
+
+	labelSearch = append(labelSearch, notEqualsLabels...)
+	labelSearch = append(labelSearch, existsKeys...)
+	labelSearch = append(labelSearch, doesNotExistKeys...)
+	return labelSelector, strings.Join(labelSearch, " and "), true, nil
```
clyang82, line 172: Does kube label selector support `or`?

skeeey: no, it does not

pkg/client/cloudevents/grpcsource/util.go
skeeey: done

clyang82 (approved):
LGTM

machi1990:
> lgtm
> 
> Once merged, I can test it against the client I am building.

Just an FYI; I've tested the PR with a code similar to this 
```

l := &corev1.NamespaceList{}

err = client.List(ctx, l, k8sClient.MatchingLabels(map[string]string{
"key": "label",
}))

```

and works as expected. Thanks @skeeey @clyang82 @qiujian16

PR #136, opened 2024-06-20
Use common SQL Parser for maestro search

clyang82:
depends on https://github.com/openshift-online/ocm-common/pull/55

refer to this test for the cases covered by sql parser: https://github.com/openshift-online/ocm-common/blob/21774a5cbeb153e0dc687530c1343c16996824e4/pkg/utils/parser/sql_parser/sql_parser_test.go


pkg/services/generic.go
```diff
@@ -156,12 +155,12 @@ func (s *sqlGenericService) buildSearch(listCtx *listContext, d *dao.GenericDao)
 	}
 
 	if isJSONBSearch(listCtx.args.Search) {
-		// focus on jsonb, ignore type = 'Single' or 'Bundle'
-		jsonbSearch, _, _ := strings.Cut(listCtx.args.Search, "and type=")
-		if err := validateJSONBSearch(jsonbSearch); err != nil {
-			return false, errors.BadRequest("failed to validate search query: %v", err)
+		parser := sql_parser.NewSQLParser()
```
ziccardi, line 158: This way, you are using all the defaults: any column is accepted, and the maximum complexity (maximum number of logic operators) is 10.

Do we want to allow the user to use any column name into the query?

If needed, you can restrict the columns that can be used in the query to a specified list of columns, and/or you can change the maximum complexity.
That can be easily done by passing the options to the `NewSQLParser` method:
```go
parser := sql_parser.NewSQLParser(
   WithValidColumns("column1", "column2", "column3"),
   WithMaximumComplexity(15),
)
```

clyang82: Thanks @ziccardi for your guidance. I think any column is accepted. our cases are about 3 or 4 complexity right now. so the default value is good for us.

qiujian16 (approved):

PR #137, opened 2024-06-20
Add e2e test for broadcast subscription type of resource status update

morvencao:
This PR adds an E2E test for scenarios where the MQTT broker doesn't support shared subscriptions, requiring the use of broadcast subscription for the resource status dispatcher.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -23,10 +23,6 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: ${{ env.GO_VERSION }}
-      - name: Setup kind
-        uses: engineerd/setup-kind@v0.5.0
-        with:
-          version: v0.17.0
       - name: install ginkgo
         run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.15.0
       - name: Test E2E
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -23,14 +23,36 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: ${{ env.GO_VERSION }}
-      - name: Setup kind
-        uses: engineerd/setup-kind@v0.5.0
+      # - name: Setup kind
```
clyang82, line 26: will it be provided by default?

morvencao: The kind binary will be installed using e2e_setup.sh (see: https://github.com/openshift-online/maestro/blob/main/test/e2e/setup/e2e_setup.sh#L20-L25)).
This action sets up a KinD cluster named **kind**, which is not needed for our purposes. Instead, we will create a KinD cluster named **maestro** with our own configuration.

qiujian16:
we need to do a rebase for this

morvencao:
rebased the code @qiujian16

clyang82 (approved):
LGTM

PR #139, opened 2024-06-21
fix maestro server resync after reconnecting.

morvencao:
this is a follow up PR for #121 to fix resync status when maestro server reconnects to the MQTT broker, also add test case for it.

@qiujian16 @clyang82 @skeeey

### DIFF
--- a/pkg/dispatcher/noop_dispatcher.go
+++ b/pkg/dispatcher/noop_dispatcher.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/openshift-online/maestro/pkg/client/cloudevents"
 	"github.com/openshift-online/maestro/pkg/dao"
+	"github.com/openshift-online/maestro/pkg/logger"
 )
 
 var _ Dispatcher = &NoopDispatcher{}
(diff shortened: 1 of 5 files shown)


pkg/dispatcher/noop_dispatcher.go
```diff
@@ -28,6 +29,36 @@ func NewNoopDispatcher(consumerDao dao.ConsumerDao, sourceClient cloudevents.Sou
...
+			consumers, err := d.consumerDao.All(ctx)
+			if err != nil {
+				log.Error(fmt.Sprintf("failed to get all consumers: %v", err))
+				continue
+			}
+
+			for _, c := range consumers {
+				consumerIDs = append(consumerIDs, c.ID)
+			}
+			if err := d.sourceClient.Resync(ctx, consumerIDs); err != nil {
```
clyang82, line 57: do we concern about the duplicated works here?

morvencao: what do you mean by duplicated works?

clyang82: after discussed offline, it should be OK since:
1. resync response will be distributed to the certain maestro instance. 
2. agent will not send the resync status repeatedly if the resync status is sent once.

clyang82 (approved):
LGTM

Issue #144, opened 2024-06-26
integration test is flaky


clyang82:
```
[unnamed-0] === RUN   TestPulseServer
[unnamed-0] {"level":"info","ts":1719336586.0897985,"logger":"fallback","caller":"v2@v2.0.0-20231030012137-0836a524e995/protocol.go:124","msg":"subscribing to topics: map[sources/maestro/consumers/cluster1/sourceevents:{1 0 false false}]"}
[unnamed-0] W0625 17:29:46.089897   12218 agentoptions.go:63] the agent broadcast topic not set, fall back to the agent events topic
[unnamed-0]     pulse_server_test.go:128: 
[unnamed-0]         Timed out after 10.017s.
[unnamed-0]         Expected success, but got an error:
[unnamed-0]             <*errors.errorString | 0xc000c52430>: 
[unnamed-0]             resource status is empty
[unnamed-0]             {
[unnamed-0]                 s: "resource status is empty",
[unnamed-0]             }
[unnamed-0] --- FAIL: TestPulseServer (14.45s)
```

morvencao:
fixed in https://github.com/openshift-online/maestro/pull/141/commits/f8cfe08a4379aae4b04b5a3a1d35faa8ad44f265#diff-7c2bc297e0c49a3705e9ed11296fe33bce3b3b3d53c991b7c15e8a43b752cd02

PR #147, opened 2024-06-27
fix flaky pulse server test.

morvencao:
fix: https://github.com/openshift-online/maestro/issues/144


clyang82 (approved):
Thanks for quick fix.

PR #154, opened 2024-07-03
Consumer deletion

yanmxa:

Note: 

Delete will remove the consumer from the storage:
1. Perform a hard delete on the consumer, the resource creation will be blocked by it.
2. Forbid consumer deletion if there are associated resources(include the marked as deleted resources).

TODO: Additional deletion options or strategies might be added in the future.

### DIFF
--- a/pkg/dao/consumer.go
+++ b/pkg/dao/consumer.go
@@ -13,7 +13,7 @@ type ConsumerDao interface {
 	Get(ctx context.Context, id string) (*api.Consumer, error)
 	Create(ctx context.Context, consumer *api.Consumer) (*api.Consumer, error)
 	Replace(ctx context.Context, consumer *api.Consumer) (*api.Consumer, error)
-	Delete(ctx context.Context, id string) error
+	Delete(ctx context.Context, id string, unscoped bool) error
 	FindByIDs(ctx context.Context, ids []string) (api.ConsumerList, error)
 	All(ctx context.Context) (api.ConsumerList, error)
 }
@@ -55,8 +55,12 @@ func (d *sqlConsumerDao) Replace(ctx context.Context, consumer *api.Consumer) (*
 	return consumer, nil
 }
 
-func (d *sqlConsumerDao) Delete(ctx context.Context, id string) error {
+func (d *sqlConsumerDao) Delete(ctx context.Context, id string, unscoped bool) error {
 	g2 := (*d.sessionFactory).New(ctx)
+	if unscoped {
+		// Unscoped is used to permanently delete the record
+		g2 = g2.Unscoped()
+	}
 	if err := g2.Omit(clause.Associations).Delete(&api.Consumer{Meta: api.Meta{ID: id}}).Error; err != nil {
 		db.MarkForRollback(ctx, err)
 		return err
--- a/pkg/dao/resource.go
+++ b/pkg/dao/resource.go
@@ -19,6 +19,7 @@ type ResourceDao interface {
 	FindBySource(ctx context.Context, source string) (api.ResourceList, error)
 	FindByConsumerName(ctx context.Context, consumerName string) (api.ResourceList, error)
 	All(ctx context.Context) (api.ResourceList, error)
+	FirstByConsumerName(ctx context.Context, name string) (api.Resource, error)
 }
 
 var _ ResourceDao = &sqlResourceDao{}
@@ -114,4 +115,11 @@ func (d *sqlResourceDao) All(ctx context.Context) (api.ResourceList, error) {
 		return nil, err
 	}
 	return resources, nil
+}
+
+func (d *sqlResourceDao) FirstByConsumerName(ctx context.Context, consumerName string) (api.Resource, error) {
+	g2 := (*d.sessionFactory).New(ctx)
+	resource := api.Resource{}
+	err := g2.Where("consumer_name = ?", consumerName).First(&resource).Error
+	return resource, err
 }
--- a/pkg/dao/resource.go
+++ b/pkg/dao/resource.go
@@ -117,7 +117,7 @@ func (d *sqlResourceDao) All(ctx context.Context) (api.ResourceList, error) {
 	return resources, nil
 }
 
-// FirstByConsumerName will take the first item of the resources. Currently leverage it to determine whether the result exists in the table.
+// FirstByConsumerName will take the first item of the resources on the consumer. it can be used to determine whether the resource exists for the consumer.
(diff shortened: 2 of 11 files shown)


yanmxa:
The e2e case might sometimes fail with the same error in this PR https://github.com/openshift-online/maestro/pull/156. It has no relation to the current changes and won't be fixed here.

pkg/dao/resource.go
```diff
@@ -115,3 +116,10 @@ func (d *sqlResourceDao) All(ctx context.Context) (api.ResourceList, error) {
 	}
 	return resources, nil
 }
+
+func (d *sqlResourceDao) FirstByConsumerName(ctx context.Context, consumerName string) (api.Resource, error) {
```
qiujian16, line 120: Why first?

pkg/services/consumer.go
```diff
@@ -63,6 +66,20 @@ func (s *sqlConsumerService) Replace(ctx context.Context, consumer *api.Consumer
...
+		return err
+	}
+	_, e := s.resourceDao.FirstByConsumerName(ctx, consumer.Name)
+	if e == nil {
+		return errors.Forbidden("Resources associated with the consumer: %s", consumer.Name)
+	}
+
+	if e != gorm.ErrRecordNotFound {
+		return handleGetError("Resource", "consumer_name", consumer.Name, e)
+	}
```
qiujian16, line 80: return error with other type

pkg/services/consumer.go
```diff
@@ -63,6 +66,20 @@ func (s *sqlConsumerService) Replace(ctx context.Context, consumer *api.Consumer
...
+
+	// e is record not found
```
qiujian16, line 82: something to consider is, we need to block creation of resource before delete consumer.

pkg/dao/resource.go
yanmxa: I just chose one(first) item to determine whether the resource exists for the consumer. Of course, we can choose to use the current [FindByConsumerName](https://github.com/openshift-online/maestro/blob/main/pkg/dao/resource.go#L20) to do that. The latter may need more memory footprint, call it will load all the resources on the consumer to the memory. What do you think of it?

pkg/services/consumer.go
yanmxa: Done.

pkg/services/consumer.go
yanmxa: Thanks @qiujian16! 

I think of two ways to block the resource creation when deleting the consumer.

**Optional 1**: Hard delete the consumer, and leverage the foreign key `fk_resources_consumers` to block creating a resource on it.
  ```bash
  maestro=# \d resources;
                           Table "public.resources"
      Column     |           Type           | Collation | Nullable | Default
  ---------------+--------------------------+-----------+----------+---------
   name          | text                     |           |          |
   id            | text                     |           | not null |
  ...
  Foreign-key constraints:
      "fk_resources_consumers" FOREIGN KEY (consumer_name) REFERENCES consumers(name) ON UPDATE RESTRICT ON DELETE RESTRICT
  ```
  It will throw an error like this to block the creation for the deleted consumer:
  ```bash
  Unable to create Resource: pq: insert or update on table "resources" violates foreign key constraint "fk_resources_consumers"
  ```

**Optional 2**: Keep deleting the consumer softly(mark it as deleted), and add validation to check the consumer when creating resources. If the consumer is not found or marked as deleted, then block the creating option. 
This doesn't seem like an elegant way, as it requires checking each resource creation operation to ensure that.


The operational 1 might be feasible for now. Do you have any other suggestions?

pkg/dao/resource.go
qiujian16: yeah, I think we need to document this with the reason. Also what if the resource is marked as deleted or deleting? do we still block?

yanmxa: Yes, I added comments for the `FirstByConsumerName`.

If a resource is marked as deleted, we will still block the deletion.

pkg/dao/consumer.go
```diff
@@ -55,8 +55,12 @@ func (d *sqlConsumerDao) Replace(ctx context.Context, consumer *api.Consumer) (*
 	return consumer, nil
 }
 
-func (d *sqlConsumerDao) Delete(ctx context.Context, id string) error {
+func (d *sqlConsumerDao) Delete(ctx context.Context, id string, unscoped bool) error {
 	g2 := (*d.sessionFactory).New(ctx)
+	if unscoped {
+		// Unscoped is used to permanently delete the record
+		g2 = g2.Unscoped()
+	}
```
clyang82, line 63: You need to lock the resource table before deleting. something like:
```
	lockOwnerID, err := s.lockFactory.NewAdvisoryLock(ctx, resource.ID, db.Resources)
	// Ensure that the transaction related to this lock always end.
	defer s.lockFactory.Unlock(ctx, lockOwnerID)
```

yanmxa: I found the advisory lock is only used for the resource payload update. Do we need to add this lock to all the write(update,create,delete) operations for the consumer or resource table?

pkg/services/consumer.go
```diff
@@ -62,8 +65,27 @@ func (s *sqlConsumerService) Replace(ctx context.Context, consumer *api.Consumer
...
+		return errors.Forbidden("Resources associated with the consumer: %s", consumer.Name)
+	}
+
+	if e != gorm.ErrRecordNotFound {
+		return errors.GeneralError("Unable to get resources by consumer: %s", e)
+	}
+
+	// e is record not found
+	if err := s.consumerDao.Delete(ctx, id, true); err != nil {
 		return handleDeleteError("Consumer", errors.GeneralError("Unable to delete consumer: %s", err))
```
morvencao, line 89: maybe add service error type at https://github.com/openshift-online/maestro/blob/main/pkg/errors/errors.go#L78-L90 and use it for consumer deletion

yanmxa: As we discussed, we can leverage the forbidden type directly.

test/e2e/pkg/consumer_test.go
```diff
@@ -66,7 +83,30 @@ var _ = Describe("Consumer", Ordered, func() {
 		})
 
 		AfterAll(func() {
-			// TODO: add the consumer deletion
+			// delete the consumer
```
clyang82, line 86: if put the delete consumer in afterall, that means it always throws error when the spec is failed before creating the consumer. for example: there is an error in beforeall. I think it is better to have a spec for deleting consumer.

yanmxa: Done.

PR #157, opened 2024-07-09
add grpc broker.

morvencao:
This PR introduces a gRPC message broker, enabling the Maestro agent to subscribe to resource specifications and publish resource statuses, similar to how an MQTT broker operates.

depends on: https://github.com/open-cluster-management-io/sdk-go/pull/68


templates/service-template.yml
```diff
@@ -109,6 +109,16 @@ parameters:
...
 
+- name: ENABLE_GRPC_BROKER
+  displayName: Enable gRPC broker
+  description: Enable gRPC broker
+  value: "false"
+
+- name: GRPC_BROKER_BINDPORT
+  displayName: gRPC Broker Bindport
+  description: gRPC broker bind port
+  value: "8090"
```
clyang82, line 120: `8091`?

templates/service-template.yml
```diff
@@ -395,19 +406,37 @@ objects:
...
       annotations:
         description: Exposes and load balances the maestro pods grpc endpoint
-        service.alpha.openshift.io/serving-cert-secret-name: maestro-tls
     spec:
       selector:
         app: maestro
       ports:
-        - port: 8090
+        - name: grpc
+          port: 8090
```
clyang82, line 418: Can we use ${GRPC_SERVER_BINDPORT}?

pkg/config/grpc_server.go
```diff
@@ -28,7 +30,9 @@ func NewGRPCServerConfig() *GRPCServerConfig {
 
 func (s *GRPCServerConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.EnableGRPCServer, "enable-grpc-server", false, "Enable gRPC server")
-	fs.StringVar(&s.BindPort, "grpc-server-bindport", "8090", "gPRC server bind port")
+	fs.StringVar(&s.ServerBindPort, "grpc-server-bindport", "8090", "gPRC server bind port")
+	fs.BoolVar(&s.EnableGRPCBroker, "enable-grpc-broker", false, "Enable gRPC broker")
```
clyang82, line 34: Can we use https://github.com/openshift-online/maestro/blob/b65178176cf59907febff9ac4388341eb0e7b801/pkg/config/message_broker.go#L31 to add `grpc` instead?

templates/service-template.yml
morvencao: After changing to `${GRPC_SERVER_BINDPORT}`, I encountered the following issue:
```
json: cannot unmarshal string into Go struct field ServicePort.spec.ports.port of type int32
```

templates/service-template.yml
morvencao: good catch, updated this to `8091`.

pkg/config/grpc_server.go
morvencao: I'm still considering if this makes sense.
Currently, the message broker type is used for the Maestro server to publish resources and subscribe to resource status, while the gRPC broker is mainly for the agent to publish status and subscribe to resource spec.

Since the Maestro server and gRPC broker run together in the same process, the Maestro server doesn't use gRPC to publish resources or subscribe to resource status. Therefore, enabling the gRPC broker for the Maestro server doesn't change the message type; it merely forwards resources to gRPC broker subscribers (work agents).

So, I'm unsure if we should change the meaning of the message type to accommodate this case.

cmd/maestro/servecmd/cmd.go
```diff
@@ -75,6 +80,10 @@ func runServer(cmd *cobra.Command, args []string) {
 	go healthcheckServer.Start()
 	go pulseServer.Start(ctx)
 	go controllersServer.Start(ctx)
+	// Start the GRPC broker if enabled
+	if grpcBroker != nil {
```
skeeey, line 84: nit: may 
```
if environments.Environment().Config.GRPCServer.EnableGRPCBroker {
grpcBroker = server.NewGRPCBroker()
  go grpcBroker.Start(ctx)
}
```

cmd/maestro/server/grpc_broker.go
```diff
@@ -0,0 +1,532 @@
...
+	found, svcErr := bkr.resourceService.Get(ctx, resourceID)
+	if svcErr != nil {
+		return nil, fmt.Errorf("failed to get resource %s, %s", resourceID, svcErr.Error())
+	}
+
+	if found.ConsumerName != consumerName {
+		return nil, fmt.Errorf("unmatched consumer name %s for resource %s", consumerName, resourceID)
+	}
+
+	specEvent, err := api.JSONMAPToCloudEvent(found.Payload)
```
skeeey, line 160: is the following part same with RESTful part? can we make a common function to handle this?

cmd/maestro/server/grpc_broker.go
```diff
@@ -0,0 +1,532 @@
...
+	return id, errChan
+}
+
+// unregister unregisters a subscriber by id
+func (bkr *GRPCBroker) unregister(id string) {
+	bkr.mu.Lock()
+	defer bkr.mu.Unlock()
+
+	close(bkr.subscribers[id].errChan)
+	delete(bkr.subscribers, id)
```
skeeey, line 247: need a log like the register

pkg/db/event_advisory_lock.go
```diff
@@ -0,0 +1,192 @@
...
+// before lock creation. This should be used only when the gRPC broker is enabled.
+type EventAdvisoryLockFactory struct {
+	connection           SessionFactory
+	lockStore            *AdvisoryLockStore
+	isConsumerSubscribed IsConsumerSubscribed
+}
+
+// NewEventAdvisoryLockFactory returns a new EventAdvisoryLockFactory with an AdvisoryLockStore
+// and a function to check if the consumer relevant to the current event is subscribed to the current gRPC broker.
+func NewEventAdvisoryLockFactory(connection SessionFactory, isConsumerSubscribed IsConsumerSubscribed) *EventAdvisoryLockFactory {
```
skeeey, line 26: can we use a new controller instead of this? I think for grpc broker, we don't need a lock for spec event, when a spec event is send, the controller broadcast the event for every grpc broker, each broker only handle event with its consumers? for service status update request, we can keep the current logic

morvencao: Last week, we discussed this with @qiujian16 and @clyang82. I understand that adding a new controller (specController without lock) means each instance must handle the event and add an <event, instance> record after the instance processes the event, similar to what we did for statusController. I'm still open to this and we can discuss it further in the future.

cmd/maestro/server/grpc_broker.go
morvencao: Yes, I created a common function (handleStatusUpdate) so we can handle the status event from either the MQTT or gRPC broker using the same logic.

cmd/maestro/server/grpc_broker.go
morvencao: added a log for unregister.

morvencao:
removed the multiple maestro server (grpc broker) instances for now in https://github.com/openshift-online/maestro/pull/157/commits/1083ec5ae48335075b8ea1868dd384722bc98ebb and added a TODO item for that, because it's not the required function for now and the implementation may impact other components.
/cc @qiujian16

cmd/maestro/environments/framework.go
```diff
@@ -173,26 +173,28 @@ func (e *Env) LoadClients() error {
...
-		if cloudEventsSourceOptions != nil {
-			e.Clients.CloudEventsSource, err = cloudevents.NewSourceClient(cloudEventsSourceOptions, e.Services.Resources())
+			cloudEventsSourceOptions, err := generic.BuildCloudEventsSourceOptions(config,
+				e.Config.MessageBroker.ClientID, e.Config.MessageBroker.SourceID)
 			if err != nil {
-				glog.Errorf("Unable to create CloudEvents Source client: %s", err.Error())
+				glog.Errorf("Unable to build cloudevent source options: %s", err.Error())
 				return err
 			}
+			if cloudEventsSourceOptions != nil {
```
skeeey, line 191: the cloudEventsSourceOptions will be nil under what cases?

cmd/maestro/servecmd/cmd.go
```diff
@@ -38,12 +38,20 @@ func runServer(cmd *cobra.Command, args []string) {
 	// Create event broadcaster to broadcast resource status update events to subscribers
 	eventBroadcaster := event.NewEventBroadcaster()
 
+	// Create the event server based on the message broker type:
+	// For gRPC, create a gRPC broker to handle resource spec and status events.
+	// For MQTT, create a Pulse server to handle resource spec and status events.
+	var eventServer server.EventServer
+	if environments.Environment().Config.MessageBroker.MessageBrokerType == "grpc" {
```
skeeey, line 45: may add log for this part

cmd/maestro/environments/framework.go
morvencao: looks like we don't need this check.

morvencao: removed this check.

cmd/maestro/servecmd/cmd.go
morvencao: added log.

skeeey:
LGTM

clyang82 (approved):
LGTM

PR #165, opened 2024-07-19
enable keepalive for grpc.

morvencao:
depends on: https://github.com/open-cluster-management-io/sdk-go/pull/70
This PR introduces keepalive parameters to the cloudevents client using gRPC. By default, these parameters are disabled. Users can optionally enable it to detect broken gRPC connections. MQTT has similar feature to detect the half-open connection.


morvencao:
closing this as the fix will be included in https://github.com/openshift-online/maestro/pull/241

PR #167, opened 2024-07-31
avoid do full list on the client part

skeeey:
This resolved the problem: when client get a full list from maestro sever, it may lead to the maestro server restart, the root cause is when the maestro server render the list result with the json, the  `json.Marshal` will need a lot of memory, this may lead to maestro server OOM, to avoid this, we should not do a full list query in the client part

### DIFF
--- a/pkg/client/cloudevents/grpcsource/pager.go
+++ b/pkg/client/cloudevents/grpcsource/pager.go
@@ -1,0 +1,104 @@
+package grpcsource
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/openshift-online/maestro/pkg/api/openapi"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+)
+
+// MaxListPageSize is the maximum size of one page, default is 400.
+// NOTE: This should be reset carefully, when increasing this value, both maestro server's memory limit
+// and the page size of resources need to be considered, if a bigger value is used, it might lead to
+// maestro server OOM.
+var MaxListPageSize int32 = 400
+
+// pageList assists client code in breaking large list queries into multiple smaller chunks of PageSize or smaller.
+func pageList(client *openapi.APIClient, search string, opts metav1.ListOptions) (*openapi.ResourceBundleList, string, error) {
+	items := []openapi.ResourceBundle{}
+
+	page, err := page(opts)
+	if err != nil {
+		return nil, "", err
+	}
+
+	limit := opts.Limit
+	if limit < 0 {
+		return nil, "", fmt.Errorf("limit cannot be less than 0")
+	}
+
+	var total int32 = 0
+	nextPage := ""
+	pageSize := pageSize(int32(limit))
+	offset := (page - 1) * pageSize
+	for {
+		klog.V(4).Infof("list works with search=%s, page=%d, size=%d", search, page, pageSize)
+		rbs, _, err := client.DefaultApi.ApiMaestroV1ResourceBundlesGet(context.Background()).
+			Search(search).
+			Page(page).
+			Size(pageSize).
+			Execute()
+		if err != nil {
+			return nil, "", err
+		}
+		klog.V(4).Infof("listed works total=%d, page=%d, size=%d", rbs.Total, rbs.Page, rbs.Size)
+
+		items = append(items, rbs.Items...)
+		total = rbs.Size + total
+		page = page + 1
+
+		if rbs.Size < pageSize {
+			// reaches the last page, stop list
+			break
+		}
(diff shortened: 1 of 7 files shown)


pkg/client/cloudevents/grpcsource/pager.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/openshift-online/maestro/pkg/api/openapi"
+	"k8s.io/klog/v2"
+)
+
+// MaxListPageSize is the maximum size of one page.
+// Default is 2000, this value is tested when the maestro server's memory limit is set to 1Gi and the
+// 2000 resources size is about 60Mi.
+// NOTE: This should be reset carefully, if a bigger value is used with the same test conditions, it will
+// lead to maestro server OOM.
+var MaxListPageSize int32 = 2000
```
morvencao, line 15: this is not configurable, right?

qiujian16: should we decide this by https://github.com/kubernetes/apimachinery/blob/95b78024e3feada7739b40426690b4f287933fd8/pkg/apis/meta/internalversion/types.go#L63?

skeeey: it's configurable, we can use `grpcsource. MaxListPageSize` increase this value if it is needed

pkg/client/cloudevents/grpcsource/pager.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		if limit == 0 {
+			// no limit, continue to list the rest of items
+			continue
+		}
+
+		if total == int32(limit) {
+			// reaches the limit, stop list
+			if (total + offset) < rbs.Total {
```
qiujian16, line 69: add some comment on this if. I think it means we have not list all yet, so return the next page number for list?

pkg/client/cloudevents/grpcsource/pager.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		klog.V(4).Infof("listed works total=%d, size=%d, page=%d", rbs.Total, rbs.Size, rbs.Page)
+
+		items = append(items, rbs.Items...)
+		total = rbs.Size + total
+		page = page + 1
+
+		if rbs.Size < pageSize {
+			// reaches the last page, stop list
+			if (total + offset) < rbs.Total {
```
qiujian16, line 55: why do we need this if? this is the last page, shouldn't we just break?

pkg/client/cloudevents/grpcsource/watcherstore.go
```diff
@@ -145,17 +142,9 @@ func (m *RESTFulAPIWatcherStore) Get(namespace, name string) (*workv1.ManifestWo
...
-
-	// TODO consider how to support configuring page
-	var page int32 = 1
-
-	var size int32 = -1
-	if opts.Limit > 0 {
-		size = int32(opts.Limit)
-	}
+// Using `metav1.NamespaceAll` to specify all namespace.
+func (m *RESTFulAPIWatcherStore) List(namespace string, opts metav1.ListOptions) (*workv1.ManifestWorkList, error) {
```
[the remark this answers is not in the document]
skeeey: this function is used by manifetwork client internal, the cs will not call it directly

pkg/client/cloudevents/grpcsource/pager.go
skeeey: yes, it is, will add some comments

pkg/client/cloudevents/grpcsource/pager.go
skeeey: yes, this judgment is needless

qiujian16 (approved):

pkg/client/cloudevents/grpcsource/pager.go
```diff
@@ -0,0 +1,104 @@
...
+	if limit < 0 {
+		return nil, "", fmt.Errorf("limit cannot be less than 0")
+	}
+
+	var total int32 = 0
+	nextPage := ""
+	pageSize := pageSize(int32(limit))
+	offset := (page - 1) * pageSize
+	for {
+		klog.V(4).Infof("list works with search=%s, page=%d, size=%d", search, page, page)
```
clyang82, line 38: the last parameter should be `pageSize`.

clyang82 (commented):
Do we need to update the example to include this feature - https://github.com/openshift-online/maestro/blob/main/examples/manifestworkclient/README.md ?

pkg/client/cloudevents/grpcsource/pager.go
skeeey: updated

PR #180, opened 2024-08-14
prevent resource update during deletion.

morvencao:
- add lock for MarkAsDeleting for resource
- return conflict for resource spec update for a deleting resource.

/assign @clyang82 @skeeey


skeeey:
LGTM

clyang82 (approved):
LGTM

Issue #140, opened 2024-06-21
Make RESTFulAPIWatcherStore work against Maestro server when JWT is enabled


machi1990:
This is to track this comment https://github.com/openshift-online/maestro/commit/cc34574cbdd7be29e552adf0cdae331b52a9b1ab#r143382998 

/cc @skeeey

PR #183, opened 2024-08-22
Support using outside context to request maestro RESTful api

skeeey:
this is used to fix https://github.com/openshift-online/maestro/issues/140


clyang82 (approved):
LGTM

PR #185, opened 2024-08-30
list resources with their event data type

skeeey:
Fix the resync error `failed to decode spec, manifests are invalid, Workload manifests should not be empty` when both bundle and single manifest codecs are enabled in maestro agent


clyang82 (approved):
LGTM

PR #188, opened 2024-09-05
ensure deletion event have different id and time.

morvencao:
This PR sets event ID and time when encoding deletion cloud event.
ref: https://issues.redhat.com/browse/ACM-14049


clyang82 (approved):
LGTM

PR #189, opened 2024-09-06
default orphan delete option for read only update strategy.

morvencao:
ref: https://issues.redhat.com/browse/ACM-13118

### DIFF
--- a/pkg/api/resource_types.go
+++ b/pkg/api/resource_types.go
@@ -185,21 +185,21 @@ func EncodeManifest(manifest, deleteOption, updateStrategy map[string]interface{
 		PropagationPolicy: workv1.DeletePropagationPolicyTypeForeground,
 	}
 
-	// set default delete option to Orphan if update strategy is ReadOnly
+	// set delete option to Orphan if update strategy is ReadOnly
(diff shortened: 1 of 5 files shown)


pkg/api/resource_types.go
```diff
@@ -193,7 +178,29 @@ func EncodeManifest(manifest, deleteOption, updateStrategy map[string]interface{
...
+	}
+
+	// set default delete option to Orphan if update strategy is ReadOnly
+	if upStrategy.Type == workv1.UpdateStrategyTypeReadOnly {
+		delOption = &workv1.DeleteOption{
+			PropagationPolicy: workv1.DeletePropagationPolicyTypeOrphan,
+		}
+	}
+
+	if len(deleteOption) != 0 {
```
clyang82, line 195: that means even if the update_strategy is readonly, the delete_option still can be configured. the requirement is if the update_strategy is readonly, the delete_option is orphan. and cannot be configured.

morvencao: updated this to force deletio option to `Orphan` when update strategy is `ReadOnly`.

clyang82 (approved):
LGTM

PR #221, opened 2024-11-28
Support az-entra auth

clyang82:
Fixed: https://issues.redhat.com/browse/ACM-15835

Refer to Azure document for how to configure workload indentity: https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster
basic steps are:
1. the serviceaccount has a new annotation injected `azure.workload.identity/client-id: "${USER_ASSIGNED_CLIENT_ID}"` [Microsoft Entra Workload ID](https://learn.microsoft.com/en-us/azure/active-directory/develop/workload-identities-overview) uses [Service Account Token Volume Projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection) (that is, a service account), to enable pods to use a Kubernetes identity.
2. ensure the deployment has the label `azure.workload.identity/use: "true"` so the pod can use workload identity.


clyang82:
Need have a separate PR to enable Flexible Server in our automation environment.

pkg/config/db.go
```diff
@@ -70,9 +83,26 @@ func (c *DatabaseConfig) ReadFiles() error {
...
+		// ARO-HCP environment variable configuration is set by the Azure workload identity webhook.
+		// Use [WorkloadIdentityCredential] directly when not using the webhook or needing more control over its configuration.
+		cred, err := azidentity.NewDefaultAzureCredential(nil)
+		if err != nil {
+			return err
+		}
+		// The access token can be expired. but the existing connections are not invalidated.
+		// TODO: how to reconnect due to the network is broken etc. Right now, gorm does not have this feature.
+		// refer to https://github.com/go-gorm/gorm/issues/5602 & https://github.com/go-gorm/gorm/pull/1721.
+		token, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{c.TokenRequestScope}})
```
morvencao, line 96: do we need to make the scope configurable?

clyang82: refer to the line
> fs.StringVar(&c.TokenRequestScope, "db-token-request-scope", c.TokenRequestScope, "Configure the token request scope for Open-Source Relational Database Management Systems in Azure")

qiujian16 (approved):

PR #229, opened 2024-12-12
ensure maestro instance ready after adding to hash ring.

morvencao:
Ref: https://issues.redhat.com/browse/ACM-16392

### DIFF
--- a/pkg/db/migrations/202401151014_add_server_instances.go
+++ b/pkg/db/migrations/202401151014_add_server_instances.go
@@ -1,6 +1,8 @@
 package migrations
 
 import (
+	"time"
+
 	"gorm.io/gorm"
 
 	"github.com/go-gormigrate/gormigrate/v2"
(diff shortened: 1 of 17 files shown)


skeeey:
LGTM

morvencao:
/assign @clyang82 
rebased the code

clyang82 (approved):
LGTM

pkg/db/migrations/202401151014_add_server_instances.go
```diff
@@ -9,6 +11,8 @@ import (
 func addServerInstances() *gormigrate.Migration {
 	type ServerInstance struct {
 		Model
+		LastHeartbeat time.Time
+		Ready         bool `gorm:"default:false"`
 	}
```
machi1990, line 17: This should have been a new migration and not an edit of an existing migration.

@morvencao @skeeey @clyang82 because of this, the migration doesn't get applied to existing deployments as it was already applied

PR #231, opened 2024-12-16
Support refresh access token

clyang82:


machi1990 (approved):
The code lgtm

This needs to be manually verified though when entra auth method is being used.

pkg/db/db_session/default.go
```diff
@@ -93,6 +94,44 @@ func (f *Default) Init(config *config.DatabaseConfig) {
...
+			return nil
+		} else if dbConfig.AuthMethod == constants.AuthMethodMicrosoftEntra {
+			if isExpired(dbConfig.Token) {
+				token, err := getAccessToken(ctx, dbConfig)
+				if err != nil {
+					return err
+				}
+				connConfig.Password = token.Token
+				dbConfig.Token = token
+			}
```
machi1990, line 110:
```suggestion
			if isExpired(dbConfig.Token) {
				token, err := getAccessToken(ctx, dbConfig)
				if err != nil {
					return err
				}
				connConfig.Password = token.Token
				dbConfig.Token = token
			} else {
			  connConfig.Password = dbConfig .Token.Token
			}
```

We need to handle the case where the token has not expired (or going to expire soon).

machi1990: Without this we might only be able to open one connection i.e one successfully `.Connect` call and all subsequent ones will fail with something like
```
2024-12-16T20:59:50.143Z	ERROR	logger/logger.go:129	  Unable to get all maestro instances: failed to connect to `host=<removed>.postgres.database.azure.com user=maestro-server database=maestro`: server error (FATAL: empty password returned by client (SQLSTATE 28P01)) 
github.com/openshift-online/maestro/pkg/logger.(*logger).log
````

File: pkg/db/db_session/default.go [hunk 2]
```diff
@@ -112,20 +151,33 @@ func waitForNotification(ctx context.Context, l *pq.Listener, callback func(id s
 		case <-time.After(10 * time.Second):
 			logger.V(10).Infof("Received no events on channel during interval. Pinging source")
 			go func() {
+				// TODO: Need to handle the error, especially in cases of network failure.
 				l.Ping()
```
machi1990, line 155: See https://github.com/openshift-online/maestro/pull/231/files#r1887602957 

We'll need to ensure we refresh the token to limit issues related to the token being expired

pkg/db/db_session/default.go
clyang82: Yes. Good catch! Thanks.

pkg/db/db_session/default.go
clyang82: Yes. that is why I added a TODO for the connection failure. Will have a separated PR for this one.

pkg/db/db_session/default.go
```diff
@@ -112,20 +153,33 @@ func waitForNotification(ctx context.Context, l *pq.Listener, callback func(id s
...
+	connstr := dbConfig.ConnectionString(true)
+	// append the password to the connection string
+	if dbConfig.AuthMethod == constants.AuthMethodPassword {
+		connstr += fmt.Sprintf(" password='%s'", dbConfig.Password)
+	} else if dbConfig.AuthMethod == constants.AuthMethodMicrosoftEntra {
+		token, err := getAccessToken(ctx, dbConfig)
+		if err != nil {
+			panic(err)
+		}
+		connstr += fmt.Sprintf(" password='%s'", token.Token)
```
machi1990, line 180: What's the behaviour of this one when `sslrootcert` is set to an empty string?

machi1990: I've noticed weird behaviour related to this e.g
```
2024-12-17T07:17:16.531Z	ERROR	logger/logger.go:129	  pq: empty password returned by client 
github.com/openshift-online/maestro/pkg/logger.(*logger).log
	/maestro/pkg/logger/logger.go:129
github.com/openshift-online/maestro/pkg/logger.(*logger).Error
	/maestro/pkg/logger/logger.go:120
github.com/openshift-online/maestro/pkg/db/db_session.newListener.func1
	/maestro/pkg/db/db_session/default.go:168
github.com/lib/pq.(*Listener).emitEvent
	/go/pkg/mod/github.com/lib/pq@v1.10.7/notify.go:797
github.com/lib/pq.(*Listener).listenerConnLoop
	/go/pkg/mod/github.com/lib/pq@v1.10.7/notify.go:817
github.com/lib/pq.(*Listener).listenerMain
	/go/pkg/mod/github.com/lib/pq@v1.10.7/notify.go:856
```	

And I think we'll need to quote the sslrootcert when creating the connection string, similar as to what we do with the password here.

pkg/db/db_session/default.go
machi1990: It has to be a fast follow because if the error occurs it'll be hard to debug.
Perhaps we could start by logging the error in this PR

pkg/config/db.go
```diff
@@ -117,13 +102,13 @@ func (c *DatabaseConfig) ConnectionStringWithName(name string, withSSL bool) str
 	var cmd string
 	if withSSL {
 		cmd = fmt.Sprintf(
-			"host=%s port=%d user=%s password='%s' dbname=%s sslmode=%s sslrootcert=%s",
-			c.Host, c.Port, c.Username, c.Password, name, c.SSLMode, c.RootCertFile,
+			"host=%s port=%d user=%s dbname=%s sslmode=%s sslrootcert=%s",
```
machi1990, line 105:
```suggestion
			"host=%s port=%d user=%s dbname=%s sslmode=%s sslrootcert='%s'",
```

it is possible for someone to configure an empty ssl root cert (in which case the system default will be used).
If we don't quote it, errors like https://github.com/openshift-online/maestro/pull/231/files#r1888013192 happens

machi1990 (commented):
@clyang82  I've done some manual testing and I've left two comments that needs to be addressed.
The one about refreshing the token for the listener side can be addressed in a fast follow MR but let's start with logging the error when pinging.
For the second comment, I left a suggestion

pkg/db/db_session/default.go
```diff
@@ -112,20 +151,33 @@ func waitForNotification(ctx context.Context, l *pq.Listener, callback func(id s
...
```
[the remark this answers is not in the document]
machi1990: It is being discussed in another comment

machi1990 (approved):

PR #232, opened 2024-12-17
purge status events

skeeey:


pkg/db/migrations/202311151856_add_consumers.go
```diff
@@ -22,7 +22,7 @@ func addConsumers() *gormigrate.Migration {
 			}
 
 			if err := CreateFK(tx, fkMigration{
-				"resources", "consumers", "consumer_name", "consumers(name)",
+				"resources", "consumers", "consumer_name", "consumers(name)", "ON DELETE RESTRICT ON UPDATE RESTRICT",
```
morvencao, line 25: from SRE comments, we can't modify existing migrations, we need to create new migration to ensure their data can migrated to new schema.

pkg/dao/mocks/instance.go
```diff
@@ -136,6 +136,10 @@ func (d *instanceDaoMock) FindByUpdatedTime(ctx context.Context, updatedTime tim
 	return instances, nil
 }
 
+func (d *instanceDaoMock) FindReadyIDs(ctx context.Context) ([]string, error) {
+	return []string{}, nil
```
morvencao, line 140: we can check if instance Ready == true from instances list, instead of return empty slice. this is useful for testing.

pkg/db/migrations/202412181141_alter_event_instances.go
```diff
@@ -0,0 +1,45 @@
...
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.AutoMigrate(&EventInstance{}); err != nil {
+				return err
+			}
+
+			return CreateFK(tx, fkMigration{
+				"event_instances", "server_instances", "instance_id", "server_instances(id)", "ON DELETE CASCADE",
+			}, fkMigration{
+				"event_instances", "status_events", "event_id", "status_events(id)", "ON DELETE CASCADE",
+			})
```
morvencao, line 27: would you alse add foreign key for `spec_event_id` on `events` table?

skeeey: no, do you need it?

pkg/dao/mocks/instance.go
skeeey: added

pkg/db/migrations/202412181141_alter_event_instances.go
skeeey: added

pkg/controllers/status_controller.go
```diff
@@ -131,3 +137,35 @@ func (sc *StatusController) add(ev api.StatusEventType, fns []StatusHandlerFunc)
...
+	// find the status events that already were dispatched to all ready instances
+	statusEventIDs, err := sc.eventInstanceDao.GetEventsAssociatedWithInstances(ctx, readyInstanceIDs)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to find handled status events from db, %v", err))
+		return
+	}
+
+	// batch delete the handled status events
+	batchSize := 500
+	for i := 0; i < len(statusEventIDs); i += batchSize {
```
clyang82, line 160: we may need to add UT to test this method. It may have problem if the length of statusEventIDs is 800.

skeeey: added

clyang82 (approved):
LGTM

PR #234, opened 2024-12-17
Fix tests

clyang82:
We have a question about why the CI passes without this fix, but the local environment does not. The reason is that in the CI environment, PostgreSQL runs in a sidecar container, making it effectively part of the same environment as the test container. In this case, no password is required for the database connection. The following explanation elaborates on this behavior.
> Note 1: The PostgreSQL image sets up trust authentication locally so you may notice a password is not required when connecting from localhost (inside the same container). However, a password will be required if connecting from a different host/container.

from https://hub.docker.com/_/postgres

### DIFF
--- a/pkg/db/db_session/test.go
+++ b/pkg/db/db_session/test.go
@@ -67,7 +67,7 @@ func (f *Test) Init(config *config.DatabaseConfig) {
 
 func initDatabase(config *config.DatabaseConfig, migrate func(db2 *gorm.DB) error) error {
 	// - Connect to `template1` DB
-	dbx, g2, cleanup := connect(config)
+	dbx, g2, cleanup := connect("template1", config)
 	defer cleanup()
 
 	for _, err := dbx.Exec(`select 1`); err != nil; {
@@ -80,7 +80,7 @@ func initDatabase(config *config.DatabaseConfig, migrate func(db2 *gorm.DB) erro
 
 func resetDB(config *config.DatabaseConfig) error {
 	// Reconnect to the default `postgres` database, so we can drop the existing db and recreate it
-	dbx, _, cleanup := connect(config)
+	dbx, _, cleanup := connect("postgres", config)
 	defer cleanup()
 
 	// Drop `all` connections to both `template1` and AMS DB, so it can be dropped and created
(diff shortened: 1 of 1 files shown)


morvencao (commented):
/lgtm

pkg/db/db_session/test.go
```diff
@@ -78,7 +80,7 @@ func initDatabase(config *config.DatabaseConfig, migrate func(db2 *gorm.DB) erro
 
 func resetDB(config *config.DatabaseConfig) error {
 	// Reconnect to the default `postgres` database, so we can drop the existing db and recreate it
```
machi1990, line 82: is this comment still valid?

pkg/db/db_session/test.go
```diff
@@ -78,7 +80,7 @@ func initDatabase(config *config.DatabaseConfig, migrate func(db2 *gorm.DB) erro
...
-	dbx, _, cleanup := connect("postgres", config)
+	dbx, _, cleanup := connect(config)
 	defer cleanup()
 
 	// Drop `all` connections to both `template1` and AMS DB, so it can be dropped and created
```
machi1990, line 86: do we still need to do this?

machi1990:
In the modified file, I see some things done to the `template1` database and I am wondering if these are things the maesto project wants to carry? e.g https://github.com/openshift-online/maestro/blob/999e92a3e21b540b880fc56d8e611a0f35a46999/pkg/db/db_session/test.go#L99

https://github.com/openshift-online/maestro/blob/999e92a3e21b540b880fc56d8e611a0f35a46999/pkg/db/db_session/test.go#L87

Plus some comments that are not actually true e.g https://github.com/openshift-online/maestro/blob/999e92a3e21b540b880fc56d8e611a0f35a46999/pkg/db/db_session/test.go#L94

clyang82:
I just realized that I should use `ConnectionStringWithName` instead of `ConnectionString` to keep the existing behaviour.
@machi1990 take a look at. Thanks.

machi1990:
> I just realized that I should use `ConnectionStringWithName` instead of `ConnectionString` to keep the existing behaviour. @machi1990 take a look at. Thanks.

I see comments referencing AMS, can we update them?

machi1990 (approved):

PR #235, opened 2024-12-18
multiple instances support for grpc broker.

morvencao:
ref: https://issues.redhat.com/browse/ACM-16035

/assign @skeeey @qiujian16 @clyang82


pkg/controllers/event_handler.go
```diff
@@ -0,0 +1,170 @@
...
+
+	eventInstances, err := h.eventInstanceDao.GetInstancesByEventID(ctx, event.ID)
+	if err != nil {
+		return fmt.Errorf("error finding processed server instances for event %s: %v", event.ID, err)
+	}
+
+	// check if all instances have processed the event
+	if !compareStrings(activeInstances, eventInstances) {
+		klog.V(10).Infof("Event %s is not processed by all instances, handled by %v, active instances %v", event.ID, eventInstances, activeInstances)
+		return nil
```
skeeey, line 137: do we need requeue here (return an error)? or there are some processes can retrigger this?

cmd/maestro/server/grpc_broker.go
```diff
@@ -469,6 +480,52 @@ func (bkr *GRPCBroker) OnStatusUpdate(ctx context.Context, eventID, resourceID s
...
+	// fast return if the event is already reconciled
+	if evt.ReconciledDate != nil {
+		return false, nil
+	}
+
+	resource, svcErr := bkr.resourceService.Get(ctx, evt.SourceID)
+	if svcErr != nil {
+		// if the resource is not found, it indicates the resource has been handled by
+		// other instances, so we can mark the event as reconciled and ignore it.
+		if svcErr.Is404() {
```
skeeey, line 501: add a log here

pkg/controllers/event_handler.go
morvencao: In this case, reconcileDate is not set, the 'SyncEvent' will trigger event be processed again?

If we return an error here, the current instance will process the event again, but the event may not be handled by other instances. so return an error doesn't help.

morvencao: I'm thinking maybe`activeInstances` is_subset_of `eventInstances` make more sense?

skeeey: > If we return an error here, the current instance will process the event again, but the event may not be handled by other instances. so return an error doesn't help.

so this means the `ReconciledDate` will be marked by the the "last" instance, right?

cmd/maestro/server/grpc_broker.go
morvencao: log added.

pkg/controllers/event_handler.go
morvencao: yes, I added some comments to explain different cases.

cmd/maestro/server/grpc_broker.go
```diff
@@ -183,32 +185,48 @@ func (bkr *GRPCBroker) Subscribe(subReq *pbv1.SubscriptionRequest, subServer pbv
...
+			return nil
 		}
 
 		klog.V(4).Infof("send the event to spec subscribers, %s", evt)
 
 		// WARNING: don't use "pbEvt, err := pb.ToProto(evt)" to convert cloudevent to protobuf
 		pbEvt := &pbv1.CloudEvent{}
 		if err = grpcprotocol.WritePBMessage(context.TODO(), binding.ToMessage(evt), pbEvt); err != nil {
-			return fmt.Errorf("failed to convert cloudevent to protobuf: %v", err)
+			klog.Errorf("failed to convert cloudevent to protobuf: %v", err)
```
skeeey, line 197: also record the resource ID

morvencao: added.

cmd/maestro/server/event_server.go
```diff
@@ -39,6 +39,9 @@ type EventServer interface {
 
 	// OnStatusUpdate handles status update events for a resource.
 	OnStatusUpdate(ctx context.Context, eventID, resourceID string) error
+
+	// returns true if the event should be processed by the current instance, otherwise false and an error.
```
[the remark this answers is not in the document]
morvencao: I need to update the comments, the error indicates if the event needs to be requeued. If the error is non-nil, the boolean value doesn't matter.
In summary, there are three possible return value combinations:
1. (true, nil) - Process the event.
2. (false, error) - An error occurred; requeue the event.
3. (false, nil) - Do not process or requeue the event.
make sense?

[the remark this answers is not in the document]
qiujian16: yes, that is clearer.

qiujian16:
lgtm

cmd/maestro/server/grpc_broker.go
```diff
@@ -442,6 +473,34 @@ func (bkr *GRPCBroker) OnStatusUpdate(ctx context.Context, eventID, resourceID s
...
+	if evt.ReconciledDate != nil {
+		return false, nil
+	}
+
+	resource, svcErr := bkr.resourceService.Get(ctx, evt.SourceID)
+	if svcErr != nil {
+		// if the resource is not found, it indicates the resource has been handled by other instances.
+		if svcErr.Is404() {
+			klog.V(10).Infof("The resource %s has been deleted, mark the event as reconciled", evt.SourceID)
+			return false, nil
```
clyang82, line 495: who will mark the event as reconciled? it should be marked by the instance which handle this resource.

morvencao: yes, the event is marked by the instance who handle this event.
This was added to handle this case:
1. resource created and then deleted, but no agent running.
2. agent start, resync the resource.
3. the resource is deleted, but the event will not marked as reconciled.
4. if handle the event again. will get 404 when get resource from db.

morvencao: @clyang82 added logic to mark event as reconciled here.

clyang82 (approved):
LGTM

PR #236, opened 2024-12-18
Recreate listener if error is occured

clyang82:
recreate the listener when the notify is nil or there is an error for Ping()
here is the document for notify is nil
> A notification channel will remain open until Unlisten is called, though connection loss might
result in some notifications being lost.  To solve this problem, Listener sends
a nil pointer over the Notify channel any time the connection is re-established
following a connection loss.  The application can get information about the
state of the underlying connection by setting an event callback in the call to
NewListener.

### DIFF
--- a/pkg/db/db_session/default.go
+++ b/pkg/db/db_session/default.go
@@ -138,7 +138,7 @@ func (f *Default) DirectDB() *sql.DB {
 	return f.db
 }
 
-func waitForNotification(ctx context.Context, l *pq.Listener, callback func(id string)) {
+func waitForNotification(ctx context.Context, l *pq.Listener, dbConfig *config.DatabaseConfig, channel string, callback func(id string)) {
 	logger := ocmlogger.NewOCMLogger(ctx)
 	for {
 		select {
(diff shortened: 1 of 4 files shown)


pkg/db/db_session/test.go
```diff
@@ -215,6 +216,8 @@ func (f *Test) ResetDB() {
 	f.wasDisconnected = true
 }
 
-func (f *Test) NewListener(ctx context.Context, channel string, callback func(id string)) {
-	newListener(ctx, f.config, channel, callback)
+func (f *Test) NewListener(ctx context.Context, channel string, callback func(id string)) *pq.Listener {
+	listener := newListener(ctx, f.config, channel)
```
skeeey, line 220: how about new the listener in the goroutine `waitForNotification`? we can ensure the listener in one goroutine, may like
```
func waitForNotification() {
   listener := newListener(ctx, f.config, channel)
   for {
     		// ...	
		// recreate the listener
		listener = newListener(ctx, dbConfig, channel)
        }
    }
}
```

pkg/db/db_session/default.go
```diff
@@ -189,12 +194,17 @@ func newListener(ctx context.Context, dbConfig *config.DatabaseConfig, channel s
...
-	waitForNotification(ctx, listener, callback)
+	return listener
 }
 
-func (f *Default) NewListener(ctx context.Context, channel string, callback func(id string)) {
-	newListener(ctx, f.config, channel, callback)
+func (f *Default) NewListener(ctx context.Context, channel string, callback func(id string)) *pq.Listener {
+	logger := ocmlogger.NewOCMLogger(ctx)
+
+	listener := newListener(ctx, f.config, channel)
```
skeeey, line 203: how about new the listener in the goroutine waitForNotification? we can maintain the listener in one place, may like
```
func waitForNotification() {
   listener := newListener(ctx, f.config, channel)
   for {
     		// ...	
		// recreate the listener
		listener = newListener(ctx, dbConfig, channel)
        }
    }
}
```

clyang82: if encapsulate the listener into `waitForNotification`, we cannot get the listener in the integration test. without the listener we cannot close the listener to generate an errListenerClosed to test the listener re-creation.

pkg/db/db_session/default.go
```diff
@@ -149,21 +149,26 @@ func waitForNotification(ctx context.Context, l *pq.Listener, callback func(id s
...
-				err := l.Ping()
-				if err != nil {
-					logger.Error(err.Error())
-				}
-			}()
+			if err := l.Ping(); err != nil {
+				logger.Infof("recreate the listener due to %s", err.Error())
+				l.Close()
+				// recreate the listener
+				l = newListener(ctx, dbConfig, channel)
```
qiujian16, line 165: will this be a problem? it seems we are changing a shared variable without lock. What happened when another is using the listener, and it is changed here?

clyang82: if reaching here, it means the listener has problem. it cannot be used anymore.

qiujian16: what happens when a caller calls the listener at the same time, will it panic?

pkg/db/db_session/default.go
skeeey: ok ...

may add some comments, the returned  listener only for test purpose, it cannot be shared by others?

pkg/db/db_session/default.go
clyang82: the listener is initialized when the maestro server is starting. https://github.com/openshift-online/maestro/blob/e28d0430f167086ea3ee590575187a1e2a8e6cce/cmd/maestro/server/controllers.go#L61
so there is no new caller to call the listener. we leverage LISTEN/NOTIFY mechanism  to sends a notification event together with an optional “payload” string to listener that has previously executed LISTEN channel.

qiujian16 (approved):

PR #251, opened 2025-02-10
Use quay.io/maestro images

clyang82:
use quay.io/maestro/... image to avoid the rate limit issue when try to pull image from docker.io


qiujian16 (approved):

clyang82:
the migration issue will be handled in a separate PR.

PR #252, opened 2025-02-12
get status hash from status cloudevent.

morvencao:
ref: https://issues.redhat.com/browse/ACM-16848
get status hash from resource status cloudevent, if not found, then it calculates the status hash by itself.


skeeey:
may also need to update the sdk-go, so that the maestro agent can sent this extension

morvencao:
@skeeey updated the go.sum to include changes in https://github.com/open-cluster-management-io/sdk-go/pull/89

clyang82 (approved):
LGTM

PR #253, opened 2025-02-12
 ARO-14939: Add initial support for opentelemetry tracing in maestro server

dinhxuanvu:


pkg/logger/operationid_middleware.go
```diff
@@ -49,3 +54,7 @@ func GetOperationID(ctx context.Context) string {
 	}
 	return ""
 }
+
+func operationIDAttribute(id string) attribute.KeyValue {
+	return attribute.String(string(OpIDKey), id)
```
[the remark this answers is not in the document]
frzifus: Y, i think i would wrap it with a `toLower` to be on the safe side.

cmd/maestro/server/routes.go
```diff
@@ -47,6 +49,10 @@ func (s *apiServer) routes() *mux.Router {
 	mainRouter := mux.NewRouter()
 	mainRouter.NotFoundHandler = http.HandlerFunc(api.SendNotFound)
 
+	if common.TracingEnabled() {
+		mainRouter.Use(otelmux.Middleware("serve"))
```
frzifus, line 53: Lets use proto + method here as spanname.

cmd/maestro/common/otlp_sdk.go
```diff
@@ -0,0 +1,99 @@
+package common
```
frzifus, line 1: Maybe we should place the initialization to a place where it can be reused by the frontend too. Wdyt?

E.g. here:  https://github.com/Azure/ARO-HCP/tree/main/internal

dinhxuanvu: I agreed but should it be somewhere else instead of "internal" package? Internal pkg usually means it is not exportable or should not be exportable.

simonpasquier: github.com/openshift-online/ocm-common might be the best option since it's already imported by https://gitlab.cee.redhat.com/service/uhc-clusters-service/ too. And it should be ok to import it in https://github.com/Azure/ARO-HCP/ too.

pkg/logger/operationid_middleware.go
```diff
@@ -17,9 +20,11 @@ const OpIDHeader OperationIDKey = "X-Operation-ID"
 func OperationIDMiddleware(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := WithOpID(r.Context())
```
simonpasquier, line 22: it looks like the middleware will always generate its own operation ID. Is there a reason why it shouldn't try to read the value from the `X-Operation-ID` header to facilitate the correlation between CS and Maestro?

I'm also surprised that Maestro doesn't seem to log the operation ID. IIUC it's only reported to the client in the HTTP response when an error occurs:

https://github.com/openshift-online/maestro/blob/adb7e9012bd01b12c10bcbb86eaf2e0bfa0126f9/pkg/handlers/framework.go#L32-L41

cmd/maestro/servecmd/cmd.go
```diff
@@ -76,6 +79,20 @@ func runServer(cmd *cobra.Command, args []string) {
...
+		tracingShutdown, err = common.InstallOpenTelemetryTracer(ctx, log)
+		if err != nil {
+			log.Error(fmt.Sprintf("Can't initialize OpenTelemetry trace provider: %v", err))
+			os.Exit(1)
+		}
+	}
+	if err != nil {
+		log.Error(fmt.Sprintf("Can't initialize OpenTelemetry trace provider: %v", err))
+		os.Exit(1)
+	}
```
simonpasquier, line 94: it seems duplicated code?
```suggestion
```

cmd/maestro/server/api_server.go
```diff
@@ -119,6 +121,16 @@ func NewAPIServer(eventBroadcaster *event.EventBroadcaster) Server {
 	)(mainHandler)
 
 	mainHandler = removeTrailingSlash(mainHandler)
+	mainHandler = traceAttributeMiddleware(mainHandler)
```
simonpasquier, line 124: Like we did for CS, we can add more attributes to the span in a follow-up PR to keep things simple.

cmd/maestro/server/api_server.go
```diff
@@ -118,6 +119,16 @@ func NewAPIServer(eventBroadcaster *event.EventBroadcaster) Server {
...
+	if common.TracingEnabled() {
+		mainHandler = otelhttp.NewHandler(mainHandler, "apiserver",
```
simonpasquier, line 124: do we need this additional middleware since we also wrap in cmd/maestro/server/routes.go?

pkg/logger/operationid_middleware.go
```diff
@@ -16,14 +20,24 @@ const OpIDHeader OperationIDKey = "X-Operation-ID"
 // Middleware wraps the given HTTP handler so that the details of the request are sent to the log.
 func OperationIDMiddleware(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx := WithOpID(r.Context())
-
-		opID, ok := ctx.Value(OpIDKey).(string)
-		if ok && len(opID) > 0 {
+		ctx := r.Context()
+		// Get operation ID from request header if existed
+		opID := r.Header.Get(string(OpIDHeader))
```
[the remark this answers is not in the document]
dinhxuanvu: This comes from your early suggestion that we should try to get `opID` from the request header if existed. If it doesn. then we generate one (original behavior) and set it to request header. That sounds like a reasonble change.

cmd/maestro/server/api_server.go
dinhxuanvu: @frzifus WDYT?

pkg/logger/operationid_middleware.go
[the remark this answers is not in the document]
dinhxuanvu: If you feel this should be in a separate PR for span attributes, then I can remove these for now.

cmd/maestro/server/api_server.go
frzifus: I think one would be fine. In the ARO Frontend we we just merged additional attributes into the trace created by gorilla-mux.

dinhxuanvu: As we discussed, we will keep this code here as it is helpful to know the path of api call. We remove the mux code on routes for now.

dinhxuanvu:
@simonpasquier @frzifus Ready for a lgtm.

frzifus (approved):

dinhxuanvu:
@geoberle @ janboll PTAL

pkg/logger/operationid_middleware.go
```diff
@@ -49,3 +55,7 @@ func GetOperationID(ctx context.Context) string {
 	}
 	return ""
 }
+
+func operationIDAttribute(id string) attribute.KeyValue {
+	return attribute.String(strings.ToLower(string(OpIDKey)), id)
```
simonpasquier, line 60: (nit) should we define our own constant for the attribute name?

cmd/maestro/servecmd/cmd.go
```diff
@@ -78,6 +80,16 @@ func runServer(cmd *cobra.Command, args []string) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 
+	tracingShutdown := func(context.Context) error { return nil }
+	log := logger.NewOCMLogger(ctx)
```
simonpasquier, line 84: why create a new logger? Also `logger.NewOCMLogger()` doesn't seem to be used anywhere else in the project :thinking:

pkg/logger/operationid_middleware.go
dinhxuanvu: The `opID` constant already exists so we can just reuse it. The only diff is it is `opID` instead of `opid` hence the `ToLower`. I didn't want to create another constant that is basically the same.

cmd/maestro/servecmd/cmd.go
dinhxuanvu: It is there. I can switch it over the zap logger which is used in this repo. I was planning to switch over when I import the ocm library in the future but I can do it now.

dinhxuanvu: Done.

frzifus (approved):

simonpasquier (commented):
/lgtm

simonpasquier (approved):

clyang82:
@dinhxuanvu Could you write a simple document to tell us how to setup opentelemetry collector and how to configure in maestro server? Add also show an example tracing data. Thanks.

cmd/maestro/common/otlp_sdk.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/openshift-online/maestro/pkg/constants"
+)
+
+// Without a specific configuration, a noop tracer is used by default.
+// At least two environment variables must be configured to enable trace export:
+//   - name: OTEL_EXPORTER_OTLP_ENDPOINT
+//     value: http(s)://<service>.<namespace>:4318
+//   - name: OTEL_TRACES_EXPORTER
+//     value: otlp
```
geoberle, line 26: nice, we can make it work easily in our deployment for ARO HCP by passing in the MDSD endpoint of the cluster

clyang82 (approved):
LGTM

Thanks @dinhxuanvu for your contribution.

PR #264, opened 2025-02-19
disable single resource api.

morvencao:
Ref: https://issues.redhat.com/browse/ACM-17947

This PR includes the following changes:
- Removed the CRUD API for the `/api/maestro/v1/resources` endpoints, which means the original `resource` API object no longer exists.
- Keep the `/api/maestro/v1/resource-bundles` GET/LIST methods for listing/getting resource bundles.
- Added a DELETE method to `/api/maestro/v1/resource-bundles/<id>` to remove a resource bundle.
- Refactored unit, integration, and e2e tests to use the CloudEvents source client for creating, updating, and deleting resources.


README.md
```diff
@@ -137,19 +137,22 @@ Authentication in the default configuration is done through the RedHat SSO, so y
...
+
+```shell
+
 ocm login --token=${OCM_ACCESS_TOKEN} --url=http://localhost:8000
 
 ```
 
-#### Get a new Resource
-This will be empty if no Resource is ever created
+#### List the resource bundle
```
clyang82, line 147: change it to `List the consumer`. in this step, the result is empty. then create a consumer and list again.

examples/manifestworkclient/client/README.md
```diff
@@ -0,0 +1,69 @@
+# Resource Bundle CURD with Manifestwork Client
+
+## Preparation
+
+1. Enable gRPC server by passing `--enable-grpc-server=true` to the maestro server start command:
```
clyang82, line 5: Do we need to enable gPRC server manually? I think we should make it enabled by default. WDYT?

morvencao: yes, we can do it.
if that mean delete the `--enable-grpc-server` flag? my concern is upgrade issue.

README.md
morvencao: done

examples/manifestworkclient/client/README.md
morvencao: done

clyang82 (approved):
LGTM

PR #278, opened 2025-03-14
upgrade ocm to reduce agent log size

skeeey:
Note: this change need to update the way that used to build grcp option


skeeey:
/cc @hchenxa 
This also fix the rosa test issue

clyang82 (approved):
LGTM

PR #294, opened 2025-04-16
using health check as liveness probe

skeeey:


templates/service-template.yml
```diff
@@ -397,11 +397,11 @@ objects:
                 memory: ${MEMORY_LIMIT}
             livenessProbe:
               httpGet:
-                path: /api/maestro
-                port: 8000
+                path: /healthcheck
+                port: 8083
                 scheme: HTTPS
               initialDelaySeconds: 15
```
clyang82, line 403: maybe use 20 as a consistent with readinessProbe.

templates/service-template.yml
```diff
@@ -397,11 +397,11 @@ objects:
                 memory: ${MEMORY_LIMIT}
             livenessProbe:
               httpGet:
-                path: /api/maestro
```
clyang82, line 400: Do we need to add 
```
httpHeaders:
                - name: User-Agent
                   value: Probe
```

templates/service-template.yml
skeeey: set the livenessProbe initialDelaySeconds greater than readinessProbe
use different value for periodSeconds

templates/service-template.yml
skeeey: removed

clyang82 (approved):
LGTM

PR #301, opened 2025-05-20
Remove Notify listener for server instance

clyang82:
fixed: https://issues.redhat.com/browse/ACM-20643

Remove notify/listener for the server_instance that is because 
- we found that sometimes the notify is sent out while the listener is not started, so that the channel message is missed.
- if the pod is in CrashLoopBackOff and gets restarted, the hash ring cannot be initialized since the status is no change.

Based on the above reasons, this PR is proposed to remove the notify/listener and use the existing loop to handle it per 5 seconds.

in this PR, also remove some unnecessary debug logs which can lead the logs overflow due to being full

### DIFF
--- a/pkg/db/advisory_locks.go
+++ b/pkg/db/advisory_locks.go
@@ -103,7 +103,6 @@ func (f *AdvisoryLockFactory) NewAdvisoryLock(ctx context.Context, id string, lo
 		return *lock.uuid, fmt.Errorf(errMsg)
 	}
 
-	log.Debugf("Locked advisory lock id=%s type=%s - owner=%s", id, lockType, *lock.uuid)
 	f.lockStore.add(*lock.uuid, lock)
 	return *lock.uuid, nil
 }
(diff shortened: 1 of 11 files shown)


pkg/db/advisory_locks.go
```diff
@@ -103,7 +103,6 @@ func (f *AdvisoryLockFactory) NewAdvisoryLock(ctx context.Context, id string, lo
 		return *lock.uuid, fmt.Errorf(errMsg)
 	}
 
-	log.Debugf("Locked advisory lock id=%s type=%s - owner=%s", id, lockType, *lock.uuid)
```
skeeey, line 106: maybe change the log level to trace instead of removing this?

clyang82: If an error occurs, it is captured and returned. If there is no error, it means the lock was successfully acquired. Do you really need to know the detailed information from this log?

skeeey: these locks are released by transaction，so there may be no errors, keeping this part will help us to check the lock status if some accident problems occurs on lock part

skeeey: And by default, we may use info level for maestro server logs, not the debug?

clyang82: in ARO HCP, they are using debug by default.

clyang82: Discussed with @skeeey offline, we agree that expose the lock/unlock count in metrics can address the current concern. Will submit another PR to refine the metrics to reflect on this idea.

qiujian16 (approved):

PR #328, opened 2025-07-24
add flags to disable spec controller and mqtt.

morvencao:
Signed-off-by: morvencao <lcao@redhat.com>

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

### DIFF
--- a/cmd/maestro/server/controllers.go
+++ b/cmd/maestro/server/controllers.go
@@ -20,6 +20,7 @@ func NewControllersServer(eventServer EventServer, eventFilter controllers.Event
 
 	// disable the spec controller if the message broker is disabled
 	if !env().Config.MessageBroker.Disable {
+		log.Debugf("Message broker is enabled, setting up kind controller manager")
 		s.KindControllerManager = controllers.NewKindControllerManager(
 			eventFilter,
 			env().Services.Events(),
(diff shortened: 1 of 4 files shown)


pkg/config/message_broker.go
```diff
@@ -25,6 +27,7 @@ func NewMessageBrokerConfig() *MessageBrokerConfig {
 }
 
 func (c *MessageBrokerConfig) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&c.Disable, "disable-message-broker", c.Disable, "Disable MQTT message broker, default is false")
```
clyang82, line 30: how about to hidden this flag?
_ = fs.CommandLine.MarkHidden("disable-message-broker") 
I do not want to expose it externally.

cmd/maestro/server/controllers.go
```diff
@@ -11,25 +11,29 @@ import (
...
+
+		s.KindControllerManager.Add(&controllers.ControllerConfig{
+			Source: "Resources",
+			Handlers: map[api.EventType][]controllers.ControllerHandlerFunc{
+				api.CreateEventType: {eventServer.OnCreate},
+				api.UpdateEventType: {eventServer.OnUpdate},
+				api.DeleteEventType: {eventServer.OnDelete},
+			},
+		})
+	}
```
skeeey, line 36: we may also add a high level log to indicate this flag setting

PR #351, opened 2025-09-10
Recreate watcher

clyang82:
ref: https://issues.redhat.com/browse/ACM-23937

Here is an example:
```
if !mgr.leadershipFlag.Raised() {
	if watcher != nil {
		watcher.Stop()
	}
	return
}

watcher, err = maestroManifestWorkClient.Watch(ctx, v1.ListOptions{})
if err != nil {
	mgr.logger.Error(ctx, "%v", err)
	return
}

```
Stop will stop the chans in the watch. When watch is called again, the old watcher will be removed and return a new one

### DIFF
--- a/pkg/client/cloudevents/grpcsource/watcherstore.go
+++ b/pkg/client/cloudevents/grpcsource/watcherstore.go
@@ -282,15 +282,7 @@ func (m *RESTFulAPIWatcherStore) registerWatcher(namespace string, labelSelector
 	defer m.Unlock()
 
 	watcher, ok := m.watchers[namespace]
-	if ok {
-		if watcher.isStopped() {
-			// delete the stopped watcher
-			delete(m.watchers, namespace)
-			watcher = nil
-			// recreate a new watcher and register it
-			watcher = newWorkWatcher(namespace, labelSelector)
-			m.watchers[namespace] = watcher
-		}
+	if ok && !watcher.isStopped() {
 		return watcher
 	}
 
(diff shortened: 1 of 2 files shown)


skeeey:
LGTM

pkg/client/cloudevents/grpcsource/watcherstore.go
```diff
@@ -283,6 +283,14 @@ func (m *RESTFulAPIWatcherStore) registerWatcher(namespace string, labelSelector
 
 	watcher, ok := m.watchers[namespace]
 	if ok {
+		if watcher.isStopped() {
+			// delete the stopped watcher
+			delete(m.watchers, namespace)
+			watcher = nil
```
qiujian16, line 289: this line seems not necessary

pkg/client/cloudevents/grpcsource/watcherstore.go
```diff
@@ -283,6 +283,14 @@ func (m *RESTFulAPIWatcherStore) registerWatcher(namespace string, labelSelector
 
 	watcher, ok := m.watchers[namespace]
 	if ok {
+		if watcher.isStopped() {
```
qiujian16, line 286: can we make it simpler?

```golang
if ok {
    if !watcher.isStopped {
        return watcher
    }
}
```
I do not think we need to explicily delete it from the map, it will be override in line 291

qiujian16 (approved):

PR #362, opened 2025-09-17
enhance the grpc client logs

skeeey:
1. using ocm-sdk logger instead of k8s logger in  grpc client
2. add more logs in critical path

### DIFF
--- a/examples/manifestworkclient/client-a/main.go
+++ b/examples/manifestworkclient/client-a/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/openshift-online/maestro/pkg/api/openapi"
 	"github.com/openshift-online/maestro/pkg/client/cloudevents/grpcsource"
+	"github.com/openshift-online/ocm-sdk-go/logging"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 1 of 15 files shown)


examples/manifestworkclient/client-a/main.go
```diff
@@ -28,7 +29,7 @@ const sourceID = "mw-client-example"
 var (
 	maestroServerAddr = flag.String("maestro-server", "https://127.0.0.1:30080", "The Maestro server address")
 	grpcServerAddr    = flag.String("grpc-server", "127.0.0.1:30090", "The GRPC server address")
-	consumerName      = flag.String("consumer-name", "", "The Consumer Name")
+	consumerName      = flag.String("consumer-name", "hcp-underlay-usw3lcao-mgmt-1", "The Consumer Name")
```
clyang82, line 32: remove `hcp-underlay-usw3lcao-mgmt-1`

clyang82 (commented):
What is different between ocm logger with k8s log? ocm logger = zap log?

examples/manifestworkclient/client-a/main.go
skeeey: removed

skeeey:
> What is different between ocm logger with k8s log? ocm logger = zap log?

no, ocm logger is from ocm sdk, the cs use

pkg/client/cloudevents/grpcsource/watch.go
```diff
@@ -18,18 +19,26 @@ type workWatcher struct {
 	done    chan struct{}
 	stopped bool
 
+	ctx    context.Context
```
morvencao, line 22: since we have conext for the watcher, can we delete `done` channel? they seem duplicated.

skeeey: `done` for  explicitly stop watch

clyang82 (approved):
LGTM

PR #373, opened 2025-09-26
Debug pipeline

clyang82:


.github/workflows/pipeline.yml
```diff
@@ -28,68 +28,55 @@ jobs:
...
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.15.0
+
+      # Running the e2e test
+      - name: e2e test
+        run: |
+          curl -fsSL -O https://github.com/knight42/krelay/releases/download/v0.1.2/kubectl-relay_v0.1.2_linux-amd64.tar.gz
+          tar -xzf kubectl-relay_v0.1.2_linux-amd64.tar.gz kubectl-relay
+          chmod a+x kubectl-relay
+          $(pwd)/kubectl-relay --kubeconfig ./svc-cluster.kubeconfig -n maestro svc/maestro 8000:8000 &
+          $(pwd)/kubectl-relay --kubeconfig ./svc-cluster.kubeconfig -n maestro svc/maestro-grpc 8090:8090 &
```
morvencao, line 78: the connection will be denied by istio sidecar.

hchenxa: should we have this annotation by default in this long running cluster? or let action to update the value and changed it back after the test done?

.github/workflows/pipeline.yml
```diff
@@ -28,68 +28,67 @@ jobs:
...
       - name: Check cluster access
-        run: kubectl get nodes
+        run: |
+          kubectl --kubeconfig ./svc-cluster.kubeconfig get pods -A -l app=maestro
+          kubectl --kubeconfig ./svc-cluster.kubeconfig get service -n maestro
+          kubectl --kubeconfig ./mgmt-cluster.kubeconfig get pods -A -l app=maestro-agent
+ 
+      - name: patch the istio
+        run: |
+          kubectl --kubeconfig ./svc-cluster.kubeconfig patch deploy/maestro -n maestro -p '{"spec":{"template":{"metadata":{"annotations":{"traffic.sidecar.istio.io/excludeInboundPorts":"8090","traffic.sidecar.istio.io/excludeOutboundPorts":"8090"}}}}}'
```
morvencao, line 66: we should also exclude `8000` port, since istio enable extra auth on the port, that's maybe why the consumer tests fail.

.github/workflows/pipeline.yml
morvencao: suggest removing the annotation after the test. We may add other tests (e.g., similar client with cluster-service) that require Istio to intercept the traffic.

hchenxa: yes, we have a steps to reset the annotations in line 91

.github/workflows/pipeline.yml
hchenxa: Thanks @morvencao , will add the 8000 and retry.

PR #515, opened 2026-04-08
Add diagnostic logging for slow healthcheck DB queries

patilsuraj767:
The healthcheck handler's DB call (instanceDao.Get) can exceed the Kubernetes probe timeout (default 1s), causing liveness/readiness probe failures. Use a detached context for the DB query so it runs to completion even after the kubelet disconnects, and log the actual query duration when the probe connection is outlasted. Skip writing the HTTP response when the client is already gone to avoid broken pipe errors.


jnpacker:
Is there an associated Jira issue for this?

PR #522, opened 2026-04-30
ARO-26559 fix(cloudevents): limit resync status hash payload to avoid MQTT 512KB packet limit

jnpacker:
- Remove Unscoped() from FindByConsumerName so soft-deleted resources are excluded from the hash list, fixing a data leak that inflated resource counts
  - Replace SDK CloudEventSourceClient.Resync() with a custom resyncConsumer() that computes real hashes for the first 2000 resources and sends empty hashes for the remainder, keeping the payload within the broker max_packet_size
  - Add unit tests covering hash split correctness, event structure, and payload size logging at 5000 resources

### DIFF
--- a/plans/2026-04-29-resync-status-hash-limit.md
+++ b/plans/2026-04-29-resync-status-hash-limit.md
@@ -1,0 +1,80 @@
+# Resync Status Hash Size Limit
+
+**Date:** 2026-04-29
+
+## Problem
+
+The MQTT broker has a `max_packet_size` of 512 KB. When `SourceClientImpl.Resync()` fires (on MQTT
+reconnect, consumer reassignment, or instance failover), it delegates to the SDK's
+`CloudEventSourceClient.Resync()`, which builds a `ResourceStatusHashList` containing one entry per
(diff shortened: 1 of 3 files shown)


plans/2026-04-29-resync-status-hash-limit.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Root cause of the elevated count
+
+A data leak is causing resources that are no longer real to persist in the database. The resource
+count per consumer should be ~400 once the leak is fixed. At 400 resources the full hash list is
+~61 KB, well within the broker limit.
+
+## Staged Changes
+
+### 1. `pkg/dao/resource.go` — Fix soft-deleted resource leak
```
machi1990, line 23: This change isn't there. 

But I'd like to know the rationale of it:

If the record is soft-deleted, and we don't retrieve it anymore, what guarantees to do we've that the deletion event has been at least delivered once and is being processed by the agent?

jnpacker:
using the ARO026559-split PR.

PR #523, opened 2026-04-30
ARO-26559 fix(cloudevents): send empty status hash list on resync to avoid MQTT 512KB packet limit

jnpacker:
- Remove Unscoped() from FindByConsumerName so soft-deleted resources are
    excluded, fixing a data leak that inflated resource counts to ~4000
  - Replace SDK CloudEventSourceClient.Resync() with resyncConsumer() that
    sends an empty ResourceStatusHashList (~20 bytes) instead of one hash
    entry per resource (~612KB at current counts), keeping the payload well
    within the broker max_packet_size
  - Add unit tests verifying the empty hash list is sent, that the agent's
    own decoder parses it correctly and triggers a full status republish,
    and that the CloudEvent structure is well-formed

### DIFF
--- a/pkg/client/cloudevents/source_client.go
+++ b/pkg/client/cloudevents/source_client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	cloudevents "github.com/cloudevents/sdk-go/v2"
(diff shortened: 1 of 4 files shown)


jnpacker:
using the ARO026559-split PR.

PR #528, opened 2026-05-05
[ARO-26755] Add metrics for the postgres notify buffered channel 

nimrodshn:
### Summary
Adds Prometheus metrics to monitor the PostgreSQL NOTIFY channel buffer depth and alert when it approaches capacity, helping detect potential notification drops before they occur.

### Problem
The `github.com/lib/pq` library uses a buffered channel with a [fixed capacity of 32 slots](https://github.com/lib/pq/blob/master/notify.go#L498) for NOTIFY messages (`pq.Listener.Notify`). When the buffer is full and new notifications arrive, messages are held back and the postgress connection is held. This can happen when:
- Database sends notifications faster than the application processes them
- Notification callback processing is slow or blocks
- Temporary processing delays create backpressure

Without visibility into the buffer depth, dropped notifications go undetected until they manifest as missing events downstream.

### Solution
Introduces two new Prometheus metrics to monitor NOTIFY channel health:

1. **`db_notify_channel_depth`** (Gauge)
   - Current number of pending notifications in the buffer
   - Labeled by channel name
   - Recorded on every notification received and during periodic health checks (10s interval)

2. **`db_notify_channel_full_total`** (Counter)
   - Increments when buffer depth exceeds capacity (32/32)
   - Labeled by channel name
   - Indicates the buffer is at saturation

### DIFF
--- a/pkg/db/db_session/default.go
+++ b/pkg/db/db_session/default.go
@@ -149,6 +149,13 @@ func waitForNotification(ctx context.Context, l *pq.Listener, dbConfig *config.D
 			if n != nil {
 				logger.V(4).Info("Received event from channel", "channel", n.Channel, "extra", n.Extra)
 				callback(n.Extra)
+				// Record notify channel depth metric
+				depth := len(l.Notify) + 1 // +1 accounts for the just-dequeued notification
+				notifyChannelDepthGauge.WithLabelValues(channel).Set(float64(depth))
+				if depth == notifyChannelCapacity {
+					logger.Info("Postgres NOTIFY channel buffer at capacity", "channel", channel, "capacity", notifyChannelCapacity)
+					notifyChannelFullCounter.WithLabelValues(channel).Inc()
+				}
 			} else {
 				// nil notification means the connection was closed
 				logger.Info("recreate the listener for channel due to the connection loss", "channel", channel)
--- a/pkg/db/db_session/metrics.go
+++ b/pkg/db/db_session/metrics.go
@@ -1,0 +1,36 @@
+package db_session
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	// Notify channel capacity in github.com/lib/pq
+	notifyChannelCapacity = 32
+)
+
+// Metric for PostgreSQL NOTIFY channel buffer depth
+var (
+	notifyChannelDepthGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: "db",
+			Name:      "notify_channel_depth",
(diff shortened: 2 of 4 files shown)


pkg/db/db_session/default.go
```diff
@@ -157,6 +164,14 @@ func waitForNotification(ctx context.Context, l *pq.Listener, dbConfig *config.D
 				l = newListener(ctx, dbConfig, channel)
 			}
 		case <-time.After(10 * time.Second):
+			// Record notify channel depth metric during periodic check
```
geoberle: this is not really a period check but a timeout handler.
len(l.Notify) will be 0 almost always

pkg/db/db_session/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+	notifyChannelDepthGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: "db",
+			Name:      "notify_channel_depth",
+			Help:      "Current depth of the PostgreSQL NOTIFY channel buffer (capacity: 32)",
+		},
+		[]string{"channel"},
+	)
+
+	notifyChannelNearFullCounter = prometheus.NewCounterVec(
```
geoberle: the purpose of this metric is to be able get insights into spikey behavior between the scrapes, right?

i wonder if we should go for "full" instead of "nearly full"
if we are constantly nearly full, we can see it on the `notify_channel_depth` gauge already even if we just scrape every X seconds.

geoberle (commented):
please also describe how you intent to define alerts on these meterics

pkg/db/db_session/default.go
nimrodshn: OK.

pkg/db/db_session/metrics.go
nimrodshn: > the purpose of this metric is to be able get insights into spikey behavior between the scrapes, right?

The purpose of this is to try to correlate 'full' channel with dropped events - if we miss an event when the channel is full, we can assume it was dropped here in the PG implementation.

nimrodshn: I can change this to 'full', sure.

nimrodshn:
@geoberle I would consider the following -

- `db_notify_channel_depth` > 25 (or something else) for saturation check.
- `rate(db_notify_channel_full_total[5m]) > 0` for when we are *actively* experiencing pressure on this channel.

geoberle:
> * dropping events

afaiu the pg library will not drop events but will be stuck on adding an event to the full bounded and buffered channel and eventually the db connection will be killed because of that

PR #538, opened 2026-05-13
fix(controllers): detect and re-publish resources with no status feedback

vkareh:
## Description

Resources can get persisted to the database but never delivered to the agent if the transport stream tears down between the DB commit and broker enqueue. Once the event is marked reconciled, there is no mechanism to detect the gap.

Add an undelivered resource detector that periodically scans for resources with empty status older than a configurable threshold (default 10 minutes). For each, it creates a new event to trigger re-publishing through the existing pipeline. Only resources with active (non-deleted) consumers are re-published.

New flag: `--undelivered-resource-threshold` (seconds, default 600, set to 0 to disable).

Fixes https://redhat.atlassian.net/browse/AROSLSRE-833

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD or tooling change

## Testing

1. Start server with a short threshold (60s) in gRPC mode
```maestro server --message-broker-type=grpc --undelivered-resource-threshold=60```

2. In another terminal, create a consumer and a stuck resource
```
  PGPASSWORD=foobar-bizz-buzz psql -h localhost -U maestro maestro -c "
  INSERT INTO consumers (id, name, created_at, updated_at)
  VALUES ('test-consumer-id', 'test-cluster', NOW(), NOW());

  INSERT INTO resources (id, name, version, source, consumer_name, type, payload, status, created_at, updated_at)
  VALUES ('stuck-resource', 'stuck-resource', 1, 'maestro', 'test-cluster', '', '{}', NULL, NOW() - INTERVAL '2 minutes', NOW());
  "
```

3. Watch server logs. Within ~2 minutes you should see:
```
    "Found resources with no status feedback, re-publishing" count=1
    "Re-published undelivered resource" resourceID="stuck-resource"
```

4. Verify event was created
```
  PGPASSWORD=foobar-bizz-buzz psql -h localhost -U maestro maestro \
    -c "SELECT source_id, event_type FROM events;"
  # Expected: stuck-resource | Create
```

- [x] Unit tests pass (`make test`)
- [ ] Integration tests pass (if applicable)
- [x] Manual verification completed

## Checklist

- [x] My code follows the project's coding conventions
- [x] I have updated documentation as needed
- [x] I have added tests that prove my fix/feature works
- [x] All new and existing tests pass

### DIFF
--- a/pkg/config/event_server.go
+++ b/pkg/config/event_server.go
@@ -13,8 +13,9 @@ const (
 
 // EventServerConfig contains the configuration for the message queue event server.
 type EventServerConfig struct {
-	SubscriptionType     string                `json:"subscription_type"`
-	ConsistentHashConfig *ConsistentHashConfig `json:"consistent_hash_config"`
+	SubscriptionType             string                `json:"subscription_type"`
+	ConsistentHashConfig         *ConsistentHashConfig `json:"consistent_hash_config"`
+	UndeliveredResourceThreshold int                   `json:"undelivered_resource_threshold"`
 }
 
 // ConsistentHashConfig contains the configuration for the consistent hashing algorithm.
@@ -27,8 +28,9 @@ type ConsistentHashConfig struct {
 // NewEventServerConfig creates a new EventServerConfig with default settings.
 func NewEventServerConfig() *EventServerConfig {
 	return &EventServerConfig{
-		SubscriptionType:     "shared",
-		ConsistentHashConfig: NewConsistentHashConfig(),
+		SubscriptionType:             "shared",
+		ConsistentHashConfig:         NewConsistentHashConfig(),
+		UndeliveredResourceThreshold: 600,
 	}
 }
 
@@ -52,6 +54,7 @@ func NewConsistentHashConfig() *ConsistentHashConfig {
 //     If subscription type is "broadcast", ConsistentHashConfig settings can be configured for the hashing algorithm.
 func (c *EventServerConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.SubscriptionType, "subscription-type", c.SubscriptionType, "Sets the subscription type for resource status updates from message broker, Options: \"shared\" (only one instance receives resource status message, MQTT feature ensures exclusivity) or \"broadcast\" (all instances receive messages, hashed to determine processing instance)")
+	fs.IntVar(&c.UndeliveredResourceThreshold, "undelivered-resource-threshold", c.UndeliveredResourceThreshold, "Seconds a resource can have no status (NULL) before being re-published to the message broker. Set to 0 to disable. Default: 600 (10 minutes)")
 	c.ConsistentHashConfig.AddFlags(fs)
 }
 
--- a/pkg/controllers/undelivered_detector.go
+++ b/pkg/controllers/undelivered_detector.go
@@ -1,0 +1,97 @@
+package controllers
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift-online/maestro/pkg/api"
+	"github.com/openshift-online/maestro/pkg/db"
+	"github.com/openshift-online/maestro/pkg/services"
+)
+
+var undeliveredResourcesRepublishedTotal = prometheus.NewCounter(
+	prometheus.CounterOpts{
+		Subsystem: specControllerMetricsSubsystem,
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: pkg/controllers/undelivered_detector.go [hunk 1]
```diff
@@ -0,0 +1,97 @@
...
+	lockOwnerID, acquired, err := d.lockFactory.NewNonBlockingLock(ctx, "maestro-undelivered-check", db.Instances)
+	defer d.lockFactory.Unlock(ctx, lockOwnerID)
+	if err != nil {
+		logger.Error(err, "Error obtaining the undelivered resource check lock")
+		return
+	}
+	if !acquired {
+		logger.V(4).Info("Another instance is checking undelivered resources, skip")
+		return
+	}
```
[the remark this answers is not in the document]
vkareh: This is the established pattern in this repo (check `healthcheck_server.go:131-133`). The advisory lock factory handles this gracefully.

pkg/controllers/undelivered_detector.go
```diff
@@ -0,0 +1,97 @@
...
+
+	if len(resources) == 0 {
+		return
+	}
+
+	logger.Info("Found resources with no status feedback, re-publishing", "count", len(resources))
+
+	for _, resource := range resources {
+		age := time.Since(resource.CreatedAt)
+		eventType := api.CreateEventType
```
swiencki, line 76: No dedupe per resource. Every tick creates another event row for the same stuck resource. That adds up fast during a broker outage.

pkg/dao/resource.go
```diff
@@ -128,6 +130,21 @@ func (d *sqlResourceDao) All(ctx context.Context) (api.ResourceList, error) {
...
 
+// FindUndelivered returns resources that have no status feedback and were created before
+// the cutoff time, but only if the resource's consumer still exists (not deleted).
+func (d *sqlResourceDao) FindUndelivered(ctx context.Context, threshold time.Duration) (api.ResourceList, error) {
+	g2 := (*d.sessionFactory).New(ctx)
+	resources := api.ResourceList{}
+	cutoff := time.Now().Add(-threshold)
+	if err := g2.Unscoped().
+		Joins("JOIN consumers ON consumers.name = resources.consumer_name AND consumers.deleted_at IS NULL").
+		Where("resources.deleted_at IS NULL AND (resources.status IS NULL OR resources.status = '{}') AND resources.created_at < ?", cutoff).
```
swiencki: This only catches resources that never got status. A lost Update on a resource that already reported status once won't show up here.

vkareh: Right, but that seems like a different class of bug. I think the focus for now should be on undelivered Create resources.

pkg/controllers/undelivered_detector.go
vkareh: Good catch. I should exclude resources that already have an unreconciled event in-flight.

pkg/config/event_server.go
```diff
@@ -52,6 +54,7 @@ func NewConsistentHashConfig() *ConsistentHashConfig {
 //     If subscription type is "broadcast", ConsistentHashConfig settings can be configured for the hashing algorithm.
 func (c *EventServerConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.SubscriptionType, "subscription-type", c.SubscriptionType, "Sets the subscription type for resource status updates from message broker, Options: \"shared\" (only one instance receives resource status message, MQTT feature ensures exclusivity) or \"broadcast\" (all instances receive messages, hashed to determine processing instance)")
+	fs.IntVar(&c.UndeliveredResourceThreshold, "undelivered-resource-threshold", c.UndeliveredResourceThreshold, "Seconds a resource can have empty status before being re-published to the message broker. Set to 0 to disable. Default: 600 (10 minutes)")
```
rafabene: > [!TIP]
> **nit** — non-blocking suggestion

**Category:** Inconsistency

The docs in `server.md` were updated to say "no status (NULL)" per CodeRabbit's feedback, but this flag help string still says "empty status". An operator running `--help` sees different wording than the docs. Consider aligning them:

```suggestion
	fs.IntVar(&c.UndeliveredResourceThreshold, "undelivered-resource-threshold", c.UndeliveredResourceThreshold, "Seconds a resource can have no status (NULL) before being re-published to the message broker. Set to 0 to disable. Default: 600 (10 minutes)")
```

PR #557, opened 2026-07-15
fix(server): stop orphaning applied resources on resource-bundle delete (ARO-28432)

jnpacker:
## Description

On a resource-bundle `DELETE`, Maestro server could remove the bundle from its own
database and return success to the client without ever propagating the delete to the
agent, leaving the applied `ManifestWork` (and everything it deploys) orphaned on the
managed cluster.

**Root cause:** `SourceClientImpl.OnCreate` — the handler invoked by the event
controller when processing a `CreateEventType` event — assumed that if the resource
was already marked as deleting (`DeletedAt` set) by the time its `Create` event was
processed, the agent could never have received it, and hard-deleted the resource from
the database with no propagation to the agent. That assumption is false whenever the
`Create` event being processed is stale or retried (e.g. re-queued after a transient
publish failure) for a resource that was already delivered to, and is being actively
managed by, the agent through another path. In production this caused Maestro to
report a delete as successful while the ManifestWork (a HostedCluster/HostedControlPlane
bundle) remained live on the management cluster, cascading into a permanently stuck
cluster deletion.

**Fix:** `OnCreate` now skips (does not hard-delete) when the resource is already
marked as deleting, deferring to the `Delete` event handler to propagate the
`delete_request` to the agent. The same `DeletedAt` guard is applied to
`GRPCBroker.OnCreate` for consistency with the MQTT source-client path, and
`GRPCBroker.OnDelete`/`OnUpdate` now treat a 404 as a no-op instead of an error to
avoid infinite event retries.

Two additional fixes were needed to validate this change end-to-end and are included
here:
- `Makefile`: the `lint` target used the golangci-lint v1 `-e` flag, which is
  incompatible with the v2 binary required now that `go.mod` declares `go 1.25.0`
  (no golangci-lint v1 release supports go1.25+).
- `charts/maestro-agent/templates/crd.yaml`: the `AppliedManifestWork` CRD's
  top-level `status` block was mis-indented under `spec`, which made Helm's
  server-side apply reject the CRD (`unknown field "spec.status"`) and blocked
  `maestro-agent` from installing on any fresh cluster.

Jira: [ARO-28432](https://redhat.atlassian.net/browse/ARO-28432)

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [x] CI/CD or tooling change (Makefile lint flag, chart CRD fix)

## Testing

- Added `TestOnCreateDoesNotOrphanResourceMarkedAsDeleting` (`test/integration/resource_test.go`),
  an integration test against the real DB/event server/MQTT broker that reproduces the exact
  ARO-28432 failure on the pre-fix code (`404 not found` after silent hard-delete) and passes
  with the fix.
- Added unit tests for `SourceClientImpl.OnCreate` (`pkg/client/cloudevents/source_client_test.go`)
  and `GRPCBroker.OnCreate/OnUpdate/OnDelete` (`cmd/maestro/server/grpc_broker_test.go`).
- `make test` — full unit suite passes.
- `make test-integration` (mqtt) — passes, excluding one pre-existing/flaky
  `TestNotificationQueueUsageMetric` confirmed unrelated (fails identically on `main`).
- `make lint` — now runs (was broken by the go1.25/golangci-lint v1-v2 mismatch); reports the
  same 69 pre-existing issues as `main`, zero new issues introduced by this change.
- `make e2e-test` — full suite passes: 40/40 specs, 0 failed.

- [x] Unit tests pass (`make test`)
- [x] Integration tests pass (if applicable)
- [x] Manual verification completed (full e2e suite: 40/40 passed)

## Checklist

- [x] My code follows the project's coding conventions
- [ ] I have updated documentation as needed
- [x] I have added tests that prove my fix/feature works
- [x] All new and existing tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale or repeated create events from incorrectly deleting resources already marked for deletion.
  * Improved handling of missing resources by safely ignoring outdated create, update, and delete events.
  * Ensured valid resource events continue to be delivered correctly.
  * Added safeguards to prevent resources from becoming orphaned during deletion workflows.

* **Chores**
  * Updated linting guidance and configuration for the required tool version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,10 +174,11 @@ verify: check-gopath verify-fmt-imports
 .PHONY: verify
 
 # Runs our linter to verify that everything is following best practices
-# Requires golangci-lint to be installed @ $(go env GOPATH)/bin/golangci-lint
+# Requires golangci-lint (v2, since go.mod requires go >= 1.25 which is only supported by
+# golangci-lint v2) to be installed @ $(go env GOPATH)/bin/golangci-lint
 # Linter is set to ignore `unused` stuff due to example being incomplete by definition
 lint:
-	$(GOLANGCI_LINT_BIN) run -e unused \
+	$(GOLANGCI_LINT_BIN) run -D unused \
 		./cmd/... \
 		./pkg/...
 .PHONY: lint
--- a/charts/maestro-agent/templates/crd.yaml
+++ b/charts/maestro-agent/templates/crd.yaml
@@ -83,10 +83,10 @@ spec:
       storage: true
       subresources:
         status: {}
-  status:
-    acceptedNames:
-      kind: ""
-      plural: ""
-    conditions: []
-    storedVersions: []
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
 {{- end }}
--- a/cmd/maestro/server/grpc_broker.go
+++ b/cmd/maestro/server/grpc_broker.go
@@ -215,17 +215,43 @@ func (bkr *GRPCBroker) Start(ctx context.Context) {
 
 // OnCreate is called by the controller when a resource is created on the maestro server.
 func (s *GRPCBroker) OnCreate(ctx context.Context, resourceID string) error {
-	evt, err := s.Get(ctx, resourceID, types.CreateRequestAction)
+	logger := klog.FromContext(ctx).WithValues("resourceID", resourceID)
+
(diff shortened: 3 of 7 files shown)


raelga (approved):
/lgtm
/approve

PR #576, opened 2026-08-12
fix: republish delete events for resources stuck deleting (HYPERFLEET-1497)

ciaranRoche:
## What

Restores healing for resources stuck soft-deleted, with a hard cap. When `MarkAsDeleting` is called for a resource whose deletion is already in flight, we now re-enqueue the delete event, throttled to at most one per `--delete-event-republish-interval` (default 60s, 0 disables) per resource. The check-then-create runs under the existing per-resource advisory lock, so the cap holds across replicas.

The stale-delete detector is adjusted to match: it now only retires delete events that are themselves older than the threshold, so a fresh healing event for a long soft-deleted resource gets a full threshold of delivery attempts instead of being retired within one detector tick.

## Why

An upstream race in the OCM SDK agent store ([sdk-go#232](https://github.com/open-cluster-management-io/sdk-go/issues/232)) can erase the DeletionTimestamp from the maestro-agent's in-memory store when a delete event lands while a status patch is in flight. The agent then recreates the deleted ManifestWork on the spoke and never reports `Deleted=true`, so the server's hard-delete never fires and cluster deletion hangs until an agent reconnect happens to trigger a spec resync. We observed 16+ hours across four e2e CI runs (ARO-28958, HYPERFLEET-1497).

Before #562 (AROSLSRE-1547) every retried delete request enqueued a fresh delete event, and each re-published `delete_request` re-stamped the DeletionTimestamp in the agent store, accidentally healing the race within one retry cycle. #562 made retries silent no-ops for a good reason (the event storm was starving the spec-event worker), but that also removed the healing. This change brings the healing back deliberately: the throttle guarantees retries can never accumulate events without bound, so the AROSLSRE-1547 fix is preserved. Worst case accumulation per stuck resource is threshold/interval unreconciled rows (60 at defaults) while its agent is disconnected.

This is a bounded mitigation while the upstream fix ([sdk-go#233](https://github.com/open-cluster-management-io/sdk-go/pull/233)) is in review, and it also heals any future agent-side loss of deletion state. One caveat to be aware of: healing is driven by the client retrying deletes, which CS destructors do. Once the agent confirms the delete the resource row is hard-deleted and republishing stops naturally.

## Testing

- Unit: `TestMarkAsDeletingRepublishesStaleDeleteEvent`, `TestMarkAsDeletingRepublishDisabled`, and the existing `TestMarkAsDeletingIsIdempotent` (unchanged assertions, now also validates the throttle window). The mock event DAO now stamps `CreatedAt` on create, mirroring gorm's autoCreateTime.
- Integration (real Postgres): `TestMarkAsDeletingRepublishThrottle` (throttle + republish after backdating the event row) and `TestReconcileStaleDeleteEventsSparesFreshHealingEvents` (detector retires a 2h-old event but spares the fresh healing event on a 2h soft-deleted resource). Existing `TestReconcileStaleDeleteEvents` passes unchanged.
- Full unit suite and mqtt integration suite green locally.

Related: HYPERFLEET-1497, ARO-28958, upstream root cause sdk-go#232 / fix sdk-go#233, same symptoms in the wild [ocm#1404](https://github.com/open-cluster-management-io/ocm/issues/1404).

Note for reviewers: this touches the same `MarkAsDeleting` early-returns as the pending HYPERFLEET-1496 logging PR (#575), whichever lands second is a trivial rebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01USPzrSgidrmV9YrxRRf5tQ

## Summary by CodeRabbit

- **New Features**
  - Added configurable republishing of delete events for resources that remain in a deleting state.
  - Delete-event republishing defaults to every 60 seconds and can be adjusted with a command-line setting.
  - Set the interval to zero to disable republishing.
- **Bug Fixes**
  - Prevented duplicate delete events during repeated deletion requests.
  - Improved stale-event reconciliation so recent republished events are preserved.

### DIFF
--- a/pkg/config/event_server_test.go
+++ b/pkg/config/event_server_test.go
@@ -25,6 +25,7 @@ func TestEventServerConfig(t *testing.T) {
 				},
 				UndeliveredResourceThreshold: 600,
 				StaleDeleteEventThreshold:    3600,
+				DeleteEventRepublishInterval: 60,
 			},
 		},
 		{
@@ -41,6 +42,7 @@ func TestEventServerConfig(t *testing.T) {
 				},
 				UndeliveredResourceThreshold: 600,
 				StaleDeleteEventThreshold:    3600,
+				DeleteEventRepublishInterval: 60,
 			},
 		},
 		{
@@ -60,6 +62,7 @@ func TestEventServerConfig(t *testing.T) {
 				},
 				UndeliveredResourceThreshold: 600,
 				StaleDeleteEventThreshold:    3600,
+				DeleteEventRepublishInterval: 60,
 			},
 		},
 	}
--- a/test/integration/resource_test.go
+++ b/test/integration/resource_test.go
@@ -584,6 +584,106 @@ func TestReconcileStaleDeleteEvents(t *testing.T) {
 	Expect(pendingDeletes()).To(Equal(0))
 }
 
+// TestMarkAsDeletingRepublishThrottle verifies that retried delete requests for a resource
+// stuck soft-deleted re-publish the delete event at most once per republish interval: a
+// re-delivered delete event re-stamps the deletion state on an agent that lost it, while
+// the throttle keeps retries from accumulating events without bound (AROSLSRE-1547).
+func TestMarkAsDeletingRepublishThrottle(t *testing.T) {
+	h, _ := test.RegisterIntegration(t)
+
+	ctx := context.Background()
+
+	consumer, err := h.CreateConsumer("cluster-" + rand.String(5))
+	Expect(err).NotTo(HaveOccurred())
+	deployName := fmt.Sprintf("nginx-%s", rand.String(5))
+	resource, err := h.CreateResource(uuid.NewString(), consumer.Name, deployName, "default", 1)
+	Expect(err).NotTo(HaveOccurred())
+
+	resourceService := h.Env().Services.Resources()
+	eventService := h.Env().Services.Events()
+
+	pendingDeletes := func() int {
+		unreconciled, svcErr := eventService.FindAllUnreconciledEvents(ctx)
+		Expect(svcErr).NotTo(HaveOccurred())
+		count := 0
+		for _, e := range unreconciled {
+			if e.SourceID == resource.ID && e.EventType == api.DeleteEventType {
+				count++
+			}
+		}
+		return count
+	}
+
+	// the first delete soft-deletes and enqueues one delete event; an immediate retry is
+	// throttled by that fresh event (default 60s interval)
+	Expect(resourceService.MarkAsDeleting(ctx, resource.ID)).NotTo(HaveOccurred())
+	Expect(resourceService.MarkAsDeleting(ctx, resource.ID)).NotTo(HaveOccurred())
+	Expect(pendingDeletes()).To(Equal(1))
+
+	// age the delete event past the republish interval
+	g2 := h.Env().Database.SessionFactory.New(ctx)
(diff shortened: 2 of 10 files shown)


test/integration/resource_test.go
```diff
@@ -584,6 +584,106 @@ func TestReconcileStaleDeleteEvents(t *testing.T) {
...
+		}
+		return count
+	}
+	Expect(pendingDeletes()).To(Equal(2))
+
+	// the detector retires only the 2h-old event; the fresh healing event survives
+	count, svcErr := eventService.ReconcileStaleDeleteEvents(ctx, time.Hour)
+	Expect(svcErr).NotTo(HaveOccurred())
+	Expect(count).To(Equal(int64(1)))
+	Expect(pendingDeletes()).To(Equal(1))
```
jnpacker, line 684: Nit (agrees with CodeRabbit's earlier comment): the final `pendingDeletes() == 1` count doesn't identify *which* event survived. A regression that reconciles the fresh healing event but spares the 2h-old one would still pass this assertion.

Consider asserting on the two events directly, e.g.:

```go
allEvents, svcErr := eventService.All(ctx)
Expect(svcErr).NotTo(HaveOccurred())
var deleteEvents api.EventList
for _, e := range allEvents {
	if e.SourceID == resource.ID && e.EventType == api.DeleteEventType {
		deleteEvents = append(deleteEvents, e)
	}
}
Expect(deleteEvents).To(HaveLen(2))
sort.Slice(deleteEvents, func(i, j int) bool {
	return deleteEvents[i].CreatedAt.Before(deleteEvents[j].CreatedAt)
})
Expect(deleteEvents[0].ReconciledDate).NotTo(BeNil())  // the 2h-old event
Expect(deleteEvents[1].ReconciledDate).To(BeNil())     // the fresh healing event
```

(needs `sort` added to imports)

pkg/config/event_server_test.go
```diff
@@ -60,6 +62,7 @@ func TestEventServerConfig(t *testing.T) {
 				},
 				UndeliveredResourceThreshold: 600,
 				StaleDeleteEventThreshold:    3600,
+				DeleteEventRepublishInterval: 60,
 			},
 		},
```
jnpacker, line 67: These cases only ever exercise the default `DeleteEventRepublishInterval: 60`; the `--delete-event-republish-interval` flag itself is never set, so a broken binding or an incorrect zero-disable value wouldn't be caught here.

Suggest adding a case that overrides it, e.g.:

```go
{
	name: "custom delete event republish interval",
	input: map[string]string{
		"delete-event-republish-interval": "0",
	},
	want: &EventServerConfig{
		SubscriptionType: "shared",
		ConsistentHashConfig: &ConsistentHashConfig{
			PartitionCount:    7,
			ReplicationFactor: 20,
			Load:              1.25,
		},
		UndeliveredResourceThreshold: 600,
		StaleDeleteEventThreshold:    3600,
		DeleteEventRepublishInterval: 0,
	},
},
```

jnpacker (commented):
Nice fix — the throttled republish + advisory-lock serialization + adjusted stale-event cutoff hangs together well, and the unit/integration test coverage is solid. Just two small test-robustness suggestions inline, nothing blocking.

One more thing outside the diff: `docs/cli/server.md`'s "Message Broker Flags" table lists `--stale-delete-event-threshold` and `--stalled-delete-warning-threshold` but is missing the new `--delete-event-republish-interval` flag — worth a follow-up line there.

jnpacker (approved):
The changes the agent suggested sounded good but are not required.

